PR #385, opened 2026-03-20
feat: add Factory Floor to Related Projects

mmichelli:
Adds Factory Floor to the Related Projects table.

A startup coaching skill combining Theory of Constraints, Customer Factory, How Brands Grow, and Marketing Strategy Discipline — helps founders find and work the bottleneck. Decision-tree routed by startup stage (pre-revenue, restart, growth, scaling).

GitHub: https://github.com/Swiftner/Factory-Floor
Install: `npx @swiftner/factory-floor`


alirezarezvani:
Thanks for the contribution and for thinking of us! We appreciate you sharing Factory Floor.

At this time, we're keeping the README focused on projects that are built specifically for or tightly integrated with the Claude Code Skills ecosystem. We're not adding links to third-party tools or external libraries to avoid scope creep in the repository.

If you ever build something that extends or integrates directly with our skills library, we'd love to see a PR for that. Thanks again for your interest!

PR #396, opened 2026-03-23
Add OpenDivination to related projects

ereid7:
## Summary
- add OpenDivination to the Related Projects table

## Why
OpenDivination is a complementary public skill and SDK rather than a competing bundle of coding skills. It seems like a reasonable fit for the related-projects section because it is installable, agent-oriented, and has a clear niche around tarot and I Ching with auditable entropy provenance.


alirezarezvani:
Thanks for sharing OpenDivination — interesting project!

We're currently keeping the Related Projects section limited to projects that are built specifically for or deeply integrated with the Claude Code Skills ecosystem. We're not adding links to third-party libraries or external tools to keep the repository focused.

If you ever build something that directly extends our skills library, we'd welcome a PR for that. Thanks for thinking of us!

Issue #546, opened 2026-04-19
Plugins silently fail to load — `"skills": "./"` in manifest rejected by Claude Code path validation


kamangacode:
## Problem

All skill plugins in this marketplace fail to load in recent versions of Claude Code. The skills never appear in Claude's available skills list, even when the plugin is correctly installed and enabled.

## Affected plugins

At least these 6 plugins are broken (probably more — any plugin using `\"skills\": \"./\"`):

- `c-level-skills` (2.1.2)
- `business-growth-skills` (2.1.2)
- `executive-mentor` (2.1.2)
- `engineering-skills` (2.2.0)
- `product-skills` (2.3.0)
- `marketing-skills` (2.1.2)

## Root cause

Each plugin's `.claude-plugin/plugin.json` contains:

\`\`\`json
\"skills\": \"./\"
\`\`\`

Claude Code's plugin loader now enforces a stricter path-traversal check and rejects `./` as escaping the plugin directory. This produces these errors in the Claude Code log:

\`\`\`
[ERROR] Skill path ./ specified in manifest but escapes plugin directory for c-level-skills
[ERROR] Skill path ./ specified in manifest but escapes plugin directory for business-growth-skills
[ERROR] Skill path ./ specified in manifest but escapes plugin directory for executive-mentor
[ERROR] Skill path ./ specified in manifest but escapes plugin directory for engineering-skills
[ERROR] Skill path ./ specified in manifest but escapes plugin directory for product-skills
[ERROR] Skill path ./ specified in manifest but escapes plugin directory for marketing-skills

[DEBUG] Plugin not available for MCP: c-level-skills@claude-code-skills - error type: path-traversal
\`\`\`

The skills load silently fails — users see the plugin as \"enabled\" in `/plugin` but none of its skills (cto-advisor, ceo-advisor, etc.) show up in the Skill list.

## Reproduction

1. `/plugin marketplace add alirezarezvani/claude-skills`
2. `/plugin install c-level-skills@claude-code-skills`
3. Restart Claude Code (fresh session, not resume)
4. Try to invoke any c-level skill — not found
5. Check the Claude Code log — see the \`path-traversal\` error above

## Environment

- Claude Code: VSCode extension (latest)
- macOS Darwin 25.3.0
- Plugins installed via marketplace auto-discovery

## Suggested fix

[...]

PR #548, opened 2026-04-19
fix(engineering-skills): move sub-skills into skills/ subdirectory

VextorHorizon:
## Problem

Skills from `engineering-skills` and `engineering-advanced-skills` do not appear in Claude Code after installation. The plugins show as enabled in `/plugin` but no skills are discoverable.

## Root Cause

Two issues must be fixed together:

**1. `plugin.json` declares `"skills": "./"` which Claude Code now rejects**

Recent versions of Claude Code added strict path validation. The `./` value is treated as a path traversal attempt and the plugin silently fails to load, so no skills appear.

**2. Sub-skills are in the plugin root instead of a `skills/` subdirectory**

Even if the manifest issue was fixed, Claude Code scans for skills inside a `skills/` subdirectory. With sub-skills sitting directly in the plugin root, none of them are discovered.

Both issues must be fixed together or skills still will not appear.

## Fix

This PR applies both fixes to `engineering-team` (engineering-skills) and `engineering` (engineering-advanced-skills):

1. Move all sub-skill folders into a `skills/` subdirectory using `git mv`
2. Update `plugin.json` to set `"skills": "skills"` instead of `"skills": "./"`

The `skills/` subdirectory convention is already documented in `CONVENTIONS.md` section 5 and is used correctly by the standalone `executive-mentor` plugin, confirming this is the intended structure.

## Affected plugins (not covered by this PR)

The same two issues exist in these other plugins and will need the same fix:

- `c-level-advisor`
- `business-growth`
- `finance`
- `marketing-skill`
- `product-team`
- `project-management`
- `ra-qm-team`

Related issue: #546


VextorHorizon:
Rebased onto latest main — there was a rename conflict on `self-improving-agent` (upstream had already moved it to the correct path via #549, so I just kept that). PR is mergeable now.

The CI failures are pre-existing repo issues unrelated to this PR: broken `commit_linter` import in tests, expired `VT_API_KEY` secret, and the security audit job needs `pull-requests: write` permission for fork PRs. Nothing in this PR triggers any of them.

alirezarezvani:
Superseded by **#593** which restructured all 9 multi-skill domain plugins into the `./skills/` layout (now in `dev`). Closing as obsolete.

Thanks for the careful diagnosis in the PR body — the analysis was correct and matches the approach #593 ultimately took.

---

alirezarezvani:
Superseded by #593 (merged), which restructured the umbrella plugins into the `./skills/` layout that Claude Code expects. The fix you proposed for `engineering-team` and `engineering` was applied as part of that consolidation. Thanks @VextorHorizon for the diagnosis — closing as already-fixed-in-dev.

---

PR #687, opened 2026-05-18
feat(engineering): add desktop-manager skill for Windows desktop automation

FreyaFujo:
## Summary

- Adds `engineering/desktop-manager/` — a new Windows-only skill for full desktop control from Claude Code via PowerShell + Win32 API, requiring no third-party tools or admin elevation
- Three Python CLI scripts (stdlib-only): `window_manager.py` (list/snap/tile/move/focus), `process_manager.py` (list/launch/kill), `desktop_snapshot.py` (save/restore named layouts)
- Win32 API reference in `references/powershell-window-api.md` covering GetWindowRect, SetWindowPos, ShowWindow, multi-monitor, DPI, virtual desktops

## Test plan

- [ ] `skill_validator.py` score: **100/100** (EXCELLENT)
- [ ] `skill_security_auditor.py --strict`: **PASS** — 0 CRITICAL, 0 HIGH
- [ ] `script_tester.py`: **3/3 PASS** — syntax, argparse, stdlib-only, help, JSON output
- [ ] `python -m compileall engineering/desktop-manager/scripts/`: clean
- [ ] SKILL.md: **180 lines** (under 500-line cap)
- [ ] Manual: run `python scripts/window_manager.py list` on Windows 10/11 to verify window enumeration
- [ ] Manual: run `python scripts/window_manager.py tile --layout 2col` to verify tiling

### DIFF
--- a/engineering/desktop-manager/SKILL.md
+++ b/engineering/desktop-manager/SKILL.md
@@ -1,0 +1,243 @@
+---
+name: "desktop-manager"
+description: "Windows desktop automation skill for Claude Code. Manages windows, processes, and layouts via PowerShell and Win32 API. Use when: user wants to tile or snap windows, arrange apps side by side, launch or kill processes, list what's open, save a desktop layout, or restore a named workspace. Trigger phrases: 'tile my windows', 'snap VS Code to the left', 'launch notepad', 'kill Chrome', 'list open windows', 'save my coding layout', 'restore my layout', 'arrange windows 2-column', 'what's running', 'close all instances of X'."
+license: MIT
+metadata:
+  version: 1.0.0
+  author: VSD Communications
+  category: engineering
+  updated: 2026-05-19
+  platform: windows
+---
+
+# Desktop Manager
+
+> Tile. Snap. Launch. Restore. Full Windows desktop control from Claude Code.
+
+PowerShell + Win32 API automation for Windows 10/11. No third-party tools. No admin rights required for standard window operations.
+
+---
+
+## Slash Commands
+
+| Command | What it does |
+|---------|-------------|
+| `/desktop:windows` | List all open windows with HWND, PID, and geometry |
+| `/desktop:tile` | Tile foreground windows in 2-column, 3-column, or 2×2 grid |
+| `/desktop:snap` | Snap a specific window to a screen position |
+| `/desktop:processes` | List, launch, or kill processes |
+| `/desktop:snapshot` | Save or restore a named window layout |
+
+---
+
+## When This Skill Activates
+
+Recognize these patterns:
+
+- "Tile my windows" / "arrange side by side" / "split screen"
+- "Snap [app] to the left / right / corner"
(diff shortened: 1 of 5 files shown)


alirezarezvani (changes requested):
Thanks for this — the skill is genuinely useful and the SKILL.md is well-structured: clear slash-command table, a good "When this activates" trigger list, a thorough Win32 API reference, and stdlib-only scripts with sensible subcommand design. A few items need to be addressed before it can merge.

## Blocking

**1. Missing `.claude-plugin/plugin.json` + no `marketplace.json` entry.**
Every skill package needs its own `engineering/desktop-manager/.claude-plugin/plugin.json` (`name`, `description`, `version`, `author`, `homepage`, `repository`, `license`, `skills`) and a registration entry in the root `.claude-plugin/marketplace.json`. This is enforced by `scripts/check_plugin_json.py --all` in CI, so the build will fail without it. Use the canonical `./`-prefixed skills form — for a single-skill plugin with `SKILL.md` at the plugin root: `"skills": ["./"]`.

**2. PowerShell injection in `process_manager.py` → `launch_app()`.**
`--path` and `--args` are escaped with `.replace("'", "\\'")`, but PowerShell single-quoted strings do **not** use backslash escaping — a literal single quote is escaped by **doubling** it (`''`). As written, an input containing `'` breaks out of the quoted string and the remainder is interpreted as PowerShell. Unlike `list`/`kill` (which run names through `_sanitize_name`), `launch --path` and `--args` are passed through with only this broken escape, so this is a real injection vector.

Minimum fix:
```python
target = path.replace("'", "''")     # PowerShell single-quote escaping
safe_args = args.replace("'", "''")
```
Better: pass arguments as a PowerShell array rather than interpolating a single string, and/or invoke via `subprocess` arg lists where possible. Please also add a one-line note in the SKILL.md "Requirements"/security section that `--path`/`--args` accept arbitrary executables by design (the caller is trusted), so the trust boundary is explicit.

**3. Test plan is unverified.**
The PR checklist lists `skill_validator.py` 100/100, `skill_security_auditor.py --strict` PASS, and `script_tester.py` 3/3 — but all boxes are unchecked. Please run them and paste the output (the security auditor in particular should be re-run after the escaping fix above). `python -m compileall engineering/desktop-manager/scripts/` should be clean too.

## Non-blocking (nice to have)

**4. Slash commands aren't wired up.** SKILL.md documents `/desktop:windows`, `/desktop:tile`, etc., but no `commands/` files are included. The repo convention is a `cs-*` agent + `/cs:*` command per skill (see `engineering/caveman`, `engineering/write-a-skill`). Either add them or drop the slash-command table to match what ships.

**5. Rebase onto current `dev`.** This branch is based on the ~220-skill era; `dev` is now well ahead. Please rebase so the marketplace counts and surrounding structure line up and the diff stays clean.

The Windows-only scope is fine (`metadata.platform: windows` is honest about it). Once items 1–3 are in I'm happy to re-review. Targeting `dev` is correct 👍

---
_Reviewed via Claude Code._

---

Issue #785, opened 2026-05-31
[IMPROVEMENT] could this project support codebuff agent?   the skills folder is ~/.agents/skills


ffdjjk:
```
~/.agents/skills
```
thank you.

alirezarezvani:
Implemented in PR #836. New `scripts/sync-codebuff-skills.py` installs all skills into Codebuff's `~/.agents/skills/` folder:

```bash
python scripts/sync-codebuff-skills.py                   # symlink all 341 skills
python scripts/sync-codebuff-skills.py --copy            # hard copy instead
python scripts/sync-codebuff-skills.py --domain engineering  # one domain only
python scripts/sync-codebuff-skills.py --dry-run         # preview
```

Skills land flat at `~/.agents/skills/<skill-name>/SKILL.md` (agentskills.io standard — no format conversion needed). Validated: 341 skills synced, each one directory deep with its `SKILL.md`. Thanks for the request!

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #748, opened 2026-05-26
[BUG] .vibe skills follow up


saifjarboui:
Hi ,
following #705 , i see that vibe. didnt. detect the. skills after installation

<img width="700" height="428" alt="Image" src="https://github.com/user-attachments/assets/7c292eab-10ed-4a66-8295-53dc152b12f3" />

it shhows 0 skills after launching 

<img width="661" height="62" alt="Image" src="https://github.com/user-attachments/assets/5aa817cf-ded2-4891-8d12-087d642ed1a1" />

soulbreak:
Hi saifjarboui,
As a workaround, you can:

Do a hard copy: `python scripts/sync-vibe-skills.py --copy`
Then either:

-  custom you skills_path to particular domain by editing `~/.vibe/config.toml` with a list of   `skill_paths = [ "~/.vibe/skills/claude-skills/engineering", "~/.vibe/skills/claude-skills/project-management"]`

Or flatten the structure:
- `cd ~/.vibe/skills && find . -maxdepth 3 -type d -exec cp -r {} ~/.vibe/skills/ \; `

In my case, importing all skills makes the menu slow, so I prefer to cherry-pick interesting skills

alirezarezvani:
Fixed in PR #836. Root cause (as @soulbreak's workaround implied): Vibe only discovers skills **one directory below** each configured skill path, but the sync script nested them three levels deep (`~/.vibe/skills/claude-skills/<domain>/<skill>/`), so the default `~/.vibe/skills` path found nothing → 0 skills.

`scripts/sync-vibe-skills.py` now defaults to a **flat layout**: every skill lands directly at `~/.vibe/skills/<skill-name>/`, which Vibe picks up out of the box with no config.toml changes. Details:

- Name collisions across domains are resolved deterministically (e.g. the second `handoff` becomes `productivity-handoff`)
- The index file is written as `claude-skills-index.json` so nothing Vibe owns gets clobbered
- The old namespaced layout is still available via `--nested`, with a printed reminder that it requires `skill_paths` entries in `~/.vibe/config.toml`

Validated: all 341 skills land at depth 1 with a `SKILL.md` each, re-runs are idempotent. After the PR merges, re-run:

```bash
python scripts/sync-vibe-skills.py --copy   # or without --copy for symlinks
```

[...]

PR #847, opened 2026-06-12
fix(sync): flatten Vibe + Hermes skill layout so skills are discovered (#748)

alirezarezvani:
## Summary

Fixes **#748** — Mistral Vibe shows **0 skills** after running `sync-vibe-skills.py`.

## Root cause (confirmed from source)

Vibe's `SkillManager` discovers skills with `Path.iterdir()` over each search path — it inspects only the **immediate children** of `~/.vibe/skills/` and checks `<child>/SKILL.md`. It does **not** recurse. ([mistral-vibe `vibe/core/skills/manager.py`](https://github.com/mistralai/mistral-vibe/blob/main/vibe/core/skills/manager.py): `for skill_dir in base.iterdir(): if not skill_dir.is_dir(): continue` → `skill_dir / "SKILL.md"`.)

The previous sync nested skills under `~/.vibe/skills/claude-skills/<domain>/<skill>/`. Vibe saw only the `claude-skills/` directory, found no `SKILL.md` directly inside it, and reported **0 skills**. (The working Gemini and Codex syncs were already flat — only Vibe + Hermes nested, which is why only they broke.)

## Fix

Flatten to **one level** with a collision-safe `claude-` name prefix:

- Skills now land at `~/.vibe/skills/claude-<skill-name>/` (domain-qualified, then numeric, on name collision).
- The index becomes a plain file (`claude-skills-index.json`) — not mistaken for a skill dir.
- The legacy nested `claude-skills/` dir is removed on sync (one-time migration).
- The on-disk dir name only needs to be unique; the **displayed** skill name still comes from `SKILL.md` frontmatter, so nothing user-facing changes except discoverability.

A flat layout is discoverable by **both** recursive and non-recursive crawlers, so it's the universally-safe shape. The **identical** `sync-hermes-skills.py` had the same nested layout and gets the same fix. (Vibe's mechanism is confirmed from source; Hermes' exact crawler wasn't independently verified, but flattening is strictly safe either way — noted in the docstring.)

## Verification

Simulating Vibe's exact `iterdir()` discovery against a fresh sync:

| Target | Before | After |
|--------|--------|-------|
| Vibe   | 0 discoverable | **339** discoverable |
| Hermes | 0 discoverable | **339** discoverable |

Both compile clean (`py_compile`); legacy `claude-skills/` dir confirmed removed; index emitted as a file.

## Notes / follow-ups
- `.codex` / `.gemini` syncs are already flat and unaffected.
- Did not regenerate the actual `~/.vibe` / `~/.hermes` trees in-repo (those are user-machine paths, not committed). This only changes the sync **scripts**.

Drafted for review.

---
_Generated by [Claude Code](https://claude.ai/code/session_01QGNMQNDb4L26JurebxCwma)_

### DIFF
--- a/scripts/sync-hermes-skills.py
+++ b/scripts/sync-hermes-skills.py
@@ -4,10 +4,21 @@ sync-hermes-skills.py — Install claude-code-skills into Hermes Agent.
 
 Hermes Agent (https://github.com/NousResearch/hermes-agent) discovers skills
 from ~/.hermes/skills/. This script creates symlinks from our repo's skill
-directories into Hermes's skill directory, preserving the category structure.
+directories into Hermes's skill directory.
 
 Both tools use the agentskills.io standard (SKILL.md with YAML frontmatter),
 so no format conversion is needed — just symlink the directories.
+
+IMPORTANT — flat layout (see issue #748): the agentskills.io discovery model
+looks for a SKILL.md inside the IMMEDIATE children of each skills search path.
+Mistral Vibe's SkillManager was confirmed to use ``Path.iterdir()`` (one level,
+no recursion); the working Gemini/Codex syncs are likewise flat. An earlier
+version of this script nested skills under
+``~/.hermes/skills/claude-skills/<domain>/<skill>/``, which a one-level crawler
+cannot see. We therefore place each skill ONE level deep with a ``claude-`` name
+prefix for namespace safety. A flat layout is discoverable by both recursive and
+non-recursive crawlers, so it is the universally-safe shape regardless of the
+exact mechanism Hermes uses.
 
 Usage:
     python scripts/sync-hermes-skills.py                   # full sync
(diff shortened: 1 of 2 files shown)


alirezarezvani:
Closing as **superseded** — `dev` already contains a more complete fix for #748.

While preparing this, I found `dev`'s `scripts/sync-vibe-skills.py` was independently updated (it's the source of the merge conflict here) with a better version of the same fix:
- **Flat layout as the default** (`assign_flat_names()` — bare name, then `<domain>-<skill>`, then numeric suffix — the same collision strategy this PR used), resolving the one-level `Path.iterdir()` discovery problem.
- A `--nested` opt-in to keep the legacy namespaced layout (this PR dropped that fallback).
- A `TOOL_NAME` hook + a new `scripts/sync-codebuff-skills.py` wrapper — which also closes #785.
- The `markdown-html` domain added to `DOMAIN_DIRS`.

That implementation is a superset of this one, so there's nothing to salvage by resolving the conflict. The root-cause analysis (Vibe's `SkillManager` uses `Path.iterdir()`, one level deep) is the same conclusion both arrived at.

Recommend #748 be verified against `dev`'s current `sync-vibe-skills.py` and closed if it checks out. Closing #847.

---

PR #862, opened 2026-06-23
feat(loop-library): vendor Forward Future Loop Library skill

alirezarezvani:
## What

Installs the **Loop Library** skill from [Forward-Future/loop-library](https://github.com/Forward-Future/loop-library) into this repo as a new top-level `loop-library/` folder.

The skill is vendored **verbatim** — `SKILL.md`, `agents/openai.yaml`, `references/discover.md`, and `references/audit.md` are byte-for-byte copies of upstream's `skills/loop-library/` (sha256-verified). The folder name and internal structure are preserved unchanged, per the install request ("do not skip or change the folder").

## What a loop is

A loop is a bounded feedback system — what to do, how to verify it, what to try next, and when to stop — not permission for endless autonomy. The skill lets an agent **discover, find, audit/repair, adapt, or design** repeatable agent loops through conversation.

## About the published loops

The individual published loops (the practitioner recipes) are **not** vendored here, because they are **not in the upstream Git tree either**. Per upstream's README they are *"intentionally not committed to GitHub"* — they live in a live, database-backed catalog and are fetched at runtime:

- <https://signals.forwardfuture.ai/loop-library/catalog.md>
- <https://signals.forwardfuture.ai/loop-library/catalog.json>

The skill reads this live catalog as the single source of truth. If unreachable, it reports that published-loop discovery is temporarily unavailable rather than serving stale/forked copies — by design. So "implementing the exact same loops" = installing the skill that serves them, which is what this PR does.

## Files

```
loop-library/
├── SKILL.md              # verbatim
├── agents/openai.yaml    # verbatim
├── references/discover.md# verbatim
├── references/audit.md   # verbatim
├── LICENSE               # MIT © Forward Future (added for redistribution compliance)
└── README.md             # provenance note (added)
```

## Notes

- Not yet wired into `.claude-plugin/marketplace.json` or `plugin.json` conventions, to keep the upstream skill folder unchanged. Happy to add a plugin wrapper as a follow-up if desired.

https://claude.ai/code/session_01KEJH6wwyaEs1d6AAca2HLR

---
_Generated by [Claude Code](https://claude.ai/code/session_01KEJH6wwyaEs1d6AAca2HLR)_


alirezarezvani:
Thanks for the thorough review. Flagging a deliberate scope decision the maintainer has confirmed: **this skill is vendored strictly verbatim from `Forward-Future/loop-library`, and the upstream content will not be modified.** The install request was explicitly "implement the exact same loops, don't change the folder," so the convention gaps you list are intentional non-goals here rather than oversights:

- **3 Python scripts** — not added. Upstream ships none; this is a prompt-only guide skill (the `references/ → SKILL.md → scripts/ → assets/` pipeline is a repo convention, not a hard requirement). Adding bespoke scripts would diverge from the verbatim copy.
- **Citations in `references/audit.md` / `discover.md`** — not added, for the same reason: editing the reference docs would change the vendored content.
- **`agents/cs-loop-library.md`, `commands/cs-loop-library.md`, `plugin.json`** — these are purely additive and don't touch the vendored files, but the maintainer opted to keep the folder as a clean upstream mirror for now. Can be added as a separate follow-up wrapper if/when we want marketplace registration; intentionally out of scope for this PR.
- **Live external dependency on `signals.forwardfuture.ai`** — this is by design and documented in the folder `README.md`. Per upstream, the published loops are *intentionally not committed to Git* and are served from the live catalog; the skill already treats catalog content as untrusted reference data and reports unavailability rather than serving stale/forked copies. We are mirroring that trust model exactly rather than re-architecting it.

The one actionable CI item — the headline counter gate (`derive_counters.py --check`) tripped by the new domain folder — is fixed in 2a17df9 (skills 345→346, domains 17→18, references 705→707 across README/CLAUDE/marketplace). CI is re-running on that commit.

---

PR #878, opened 2026-07-01
Add excalidraw-skill — hand-drawn diagrams for Claude

xiaoshuai1024:
Hi! This PR adds [excalidraw-skill](https://github.com/xiaoshuai1024/excalidraw-skill) to the list.

**What it is:** An agent skill that renders 64 types of hand-drawn Excalidraw diagrams (UML, ER, sequence, architecture) to editable SVG/PNG using the real Excalidraw engine. Works in Claude Code, Cursor, ZCode and 70+ agents.

I'm the author of this project and think it fits here. Happy to move the entry to a more appropriate section if needed, or close if it's out of scope. Thanks for maintaining this list!

(Reopening — previous PR #877 was auto-closed because it targeted main; retargeting to dev per the contributing guide.)


alirezarezvani:
Thanks for the submission, @xiaoshuai1024 — and for taking the time to retarget it to `dev` per the contributing guide. The hand-drawn diagram idea is genuinely neat.

Unfortunately this one isn't a fit for the repo as it stands, so I'm going to decline it. A couple of reasons:

1. **This isn't an "awesome list" of external links.** Every skill here lives *in-tree* as a full package — `SKILL.md` + `scripts/` + `references/` + `assets/` — and is registered in `.claude-plugin/marketplace.json`. We don't add entries that point to an external repo, so a README link isn't the mechanism for contributing a skill.

2. **The Excalidraw-engine approach conflicts with our design constraints.** Skills here are **stdlib-only Python, self-contained, with no heavy runtime dependencies and no LLM/ML calls** (see `CLAUDE.md` → Design Principles). Rendering via the real Excalidraw engine pulls in a Node/browser-canvas stack, which runs against the portability rules we hold every skill to. There's also existing overlap — the `markdown-html/` skills and the Mermaid integration already cover diagram rendering.

If you'd like to bring a hand-drawn-diagram capability into the library, the path would be a **native in-tree skill** that follows the package contract above and stays within the stdlib-only constraint — happy to point you at `engineering/write-a-skill` and an existing skill folder as a template if you want to attempt that.

Really appreciate the interest in the project — closing this for now. 🙏

---

Issue #897, opened 2026-07-06
`convert.sh --tool cursor` scans `.gemini` skills and prints many "Skipping invalid frontmatter" messages


zhao45332:
Duplicate of #896 
Hi,

I was trying to generate Cursor rules by following the documentation:

```bash
./scripts/convert.sh --tool cursor
```

The script starts normally, but it scans the `.gemini/skills` directory as well and prints a large number of messages like:

```text
[!!] Skipping invalid frontmatter: ./.gemini/skills/README/SKILL.md
[!!] Skipping invalid frontmatter: ./.gemini/skills/TEMPLATE/SKILL.md
[!!] Skipping invalid frontmatter: ./.gemini/skills/a11y-audit/SKILL.md
[!!] Skipping invalid frontmatter: ./.gemini/skills/ab-test-setup/SKILL.md
...
```

I understand these may not be actual errors, but it is a bit confusing because the command is specifically converting for Cursor.

My questions are:

1. Is it expected that `convert.sh --tool cursor` scans the `.gemini` directory?
2. Should these "Skipping invalid frontmatter" messages be ignored?
3. Would it make sense for the script to skip incompatible directories automatically (or print a summary instead), so users don't think something has gone wrong?

I'm using the latest version of the repository on Windows with Git Bash.

Thanks for the great project!

Issue #885, opened 2026-07-02
si plugin: skills named 'status' and 'review' shadow Claude Code built-in /status and /review


sdkkds:
## Summary

Two skills in the `self-improving-agent` (`si`) plugin use bare `name:` values that collide with Claude Code **built-in** slash commands, causing the plugin skill to shadow the built-in when the user types the bare command:

| Skill file | `name:` | Collides with built-in |
|-----------|---------|------------------------|
| `skills/status/SKILL.md` | `status` | `/status` |
| `skills/review/SKILL.md`  | `review` | `/review` |

## Reproduction

1. Install `self-improving-agent@claude-code-skills`.
2. Type `/status` (intending the built-in account/model/connectivity dashboard).
3. Claude Code routes to the plugin skill `si:status` (memory dashboard) instead of the built-in. Same for `/review` → `si:review`.

Confirmed on Claude Code (Windows, Opus 4.8). The client's slash resolver matches the plugin skill's bare leaf name (`status`) against the built-in namespace and the plugin wins. Usage counters showed the skill firing under both `self-improving-agent:status` and `si:status`, confirming loose name matching.

## Impact

Users lose easy access to the built-in `/status` and `/review` commands after installing the plugin. There is no per-skill disable in stock Claude Code — the only user workaround is editing `name:` in the plugin cache (overwritten on update) or disabling the whole plugin (loses `/si:promote`, `/si:extract`, `/si:remember`).

## Suggested fix

Rename the two skills to namespace-safe values that don't shadow built-ins, e.g.:

- `status` → `memory-status`  (invoked as `/si:memory-status`)
- `review` → `memory-review`  (invoked as `/si:memory-review`)

Update the `plugin.json` description, both `CLAUDE.md` command lists, and each `SKILL.md` usage block accordingly. General guidance worth adopting: **plugin skill `name:` values should never equal a reserved built-in command word** (`status`, `review`, `config`, `cost`, `model`, `mcp`, `doctor`, `help`, etc.).

## Environment

- Claude Code on Windows 11, model Opus 4.8
- Plugin `self-improving-agent` v2.9.0 via `claude-code-skills` marketplace

AnnasMazhar:
This issue has already been fixed on the `dev` branch.

[...]

PR #906, opened 2026-07-08
fix: convert.sh scans tool-sync dirs; si plugin status/review shadow built-ins

alirezarezvani:
## Summary

Fixes two reported bugs:

1. **#897** (duplicate of #896) — `./scripts/convert.sh --tool cursor` scanned `.gemini/skills/*` (and would similarly hit `.claude`, `.codex`, `.hermes`, `.vibe`) and printed a wall of `[!!] Skipping invalid frontmatter` warnings. Those directories are generated/symlinked mirrors of the canonical skill folders for other tool integrations, not source-of-truth skills. On this Linux checkout the mirrors are real symlinks so `find -type f` already excludes them — but on platforms where git materializes symlinks as plain text files (e.g. Git Bash on Windows, as the reporter used), each mirrored `SKILL.md` becomes a regular file containing just the link-target path, which then fails frontmatter parsing and triggers the warning for every skill. Fixed by explicitly excluding `.claude`, `.codex`, `.codex-plugin`, `.gemini`, `.hermes`, `.vibe`, and `docs` from `convert.sh`'s `find`, matching the exclusion list already used by `scripts/derive_counters.py` and `scripts/review-new-skills.sh`.

2. **#885** — the `self-improving-agent` (`si`) plugin shipped skills named bare `status` and `review`, whose `name:` frontmatter collided with Claude Code's built-in `/status` and `/review` commands, causing the plugin to shadow the built-ins. Renamed the skill folders and `name:` fields to `memory-status` / `memory-review` (invoked as `/si:memory-status` / `/si:memory-review`), and updated every in-plugin reference (`CLAUDE.md`, `README.md`, `agents/memory-analyst.md`, `hooks/error-capture.sh`, both `reference/*.md` files, `settings.json`, `.claude-plugin/plugin.json`).

As a side effect, this also removes a latent flat-namespace collision in the `.codex`/`.gemini` symlink sync: `self-improving-agent`'s old `review` skill folder shared a bare name with `engineering-team/playwright-pro/skills/review`, so only one of the two was ever reachable via those tool-sync mirrors. After the rename, both resolve correctly.

## Checklist

- [x] Target branch is `dev`
- [x] Scripts (if any) run with `--help` without errors
- [x] No vendor-locked dependencies without open-source fallback
- [x] Follows existing directory structure (`domain/skill-name/SKILL.md`)

## Type of Change

- [x] Bug fix

## Testing

- `bash -n scripts/convert.sh` — syntax OK
- Ran `./scripts/convert.sh --tool cursor --out <tmp>` end-to-end: no `.gemini`/`.claude`/etc. noise, 353 skills converted, 1 legitimately skipped (unrelated pre-existing frontmatter issue elsewhere in the tree).
- `python3 scripts/check_paths.py <changed files>` — 0 unresolvable path references.
- `python3 scripts/check_plugin_json.py engineering-team/self-improving-agent/.claude-plugin/plugin.json` — OK.
- Verified `grep -rn "si:status\|si:review"` returns nothing left in `engineering-team/self-improving-agent/`.
- Deliberately left the `.codex`/`.gemini`/`docs/` generated-artifact drift (unrelated pre-existing skills never synced) out of this diff to keep the fix focused; those directories are produced by separate sync/doc-generation scripts and regenerating them wholesale is out of scope for this bug fix.

---
_Generated by [Claude Code](https://claude.ai/code/session_017VawvkpEUUx7mRvVsK8rBR)_


alirezarezvani:
The `CI Quality Gate` failure (`derive_counters.py --check`) is pre-existing and unrelated to this change — it also fails on `dev` HEAD (in fact with a larger drift: 359 derived skills vs the 355 documented there, vs 357 on this branch). This PR only renames two skill folders and doesn't change the skill count, so it doesn't introduce or worsen the drift. Leaving the counter-sync fix out of scope for this PR to keep the diff focused on #897/#885; happy to file a separate issue/PR for the counter refresh if useful.

---

alirezarezvani:
Good catch — pushed `0fbd70f` updating `.claude-plugin/marketplace.json`'s `self-improving-agent` entry to reference `/si:memory-review` / `/si:memory-status` and bumped its version to 2.9.1 to match the plugin manifest. Confirmed no other stale `/si:review`/`/si:status` references remain outside the generated tool-sync mirrors (`.codex`, `.gemini`, `docs/`) that are intentionally out of scope here.

---

alirezarezvani:
Good catch, confirmed — the rename did leave dangling symlinks (`.codex/skills/review`, `.gemini/skills/review/SKILL.md`) and stale index/manifest entries, since those mirrors are committed, not purely build-time artifacts. Pushed `d82f554`:

- Repointed the dangling `review` symlinks to `engineering-team/playwright-pro/skills/review` — the plugin that now legitimately owns the bare `review` name (previously shadowed by self-improving-agent's same-named skill, which is itself the flat-namespace-collision symptom of the bug this PR fixes).
- Added `memory-review`/`memory-status` symlinks + index entries for `.codex` and `.gemini`.
- Regenerated the `engineering-team` agent-harness manifest via `harness_manifest_builder.py --domain engineering-team`.

I intentionally did this by hand rather than a full `sync-codex-skills.py`/`sync-gemini-skills.py` re-run: both scripts are significantly behind on unrelated pre-existing drift (skills added in earlier merges that were never synced, e.g. `agent-harness`, `minimalist`, `strict-api`), and a full regen also non-deterministically flips a couple of unrelated bare-name collisions (`status` between `autoresearch-agent`/`agenthub`, `run` likewise) depending on filesystem walk order. Neither is something this PR should fix. Verified `.hermes`/`.vibe` don't need changes — self-improving-agent is mirrored there only as a single symlink to the plugin's main skill, not per-sub-skill, so the rename doesn't touch them. Re-ran `check_paths.py --all` clean (0 unresolvable references) and validated both index JSONs parse and their counts match actual entries.

---

alirezarezvani:
Thanks — pushed `2c2df8d` adding the missing `memory-status` entry to `.gemini/skills-index.json` (mirroring `memory-review`) and bumped `total_skills` 419 → 420 to match the actual entry count. Confirmed the `">-"` folded-scalar description issue is pre-existing in `sync-gemini-skills.py`'s frontmatter parser (unrelated to this PR — reproduces on unrelated entries like `migrate`/`report` too) and left it out of scope.

---

Issue #14, opened 2025-11-05
Create agent structure template and creation guide


alirezarezvani:
## Summary

Standardize agent creation process with template and comprehensive documentation.

## Context

This is Phase 2.4 of the Skill-Agent Integration implementation plan. This template and guide ensure consistency across all future agent development.

Reference: `documentation/implementation/implementation-plan-november-2025.md` (Phase 2.4)

### Agent Template
- [ ] Create `templates/agent-template.md` with complete structure
- [ ] Include YAML frontmatter structure with all required fields:
  - [ ] name (cs-* prefix)
  - [ ] description (clear, concise)
  - [ ] skills (reference to skill folder)
  - [ ] domain (marketing, c-level, product, etc.)
  - [ ] model (sonnet, opus, haiku)
  - [ ] tools (Read, Write, Bash, Grep, Glob)
- [ ] Document required sections:
  - [ ] Purpose (what the agent does)
  - [ ] Skill Integration (relative paths, tool examples)
  - [ ] Workflows (step-by-step processes)
  - [ ] Integration Examples (concrete code examples)
  - [ ] Success Metrics (how to measure effectiveness)
  - [ ] Related Agents (cross-references)
  - [ ] References (documentation links)
- [ ] Add placeholder content for each section
- [ ] Include comments explaining each section's purpose

### Agent Creation Guide
- [ ] Create `documentation/AGENT_CREATION_GUIDE.md`
- [ ] Document step-by-step agent creation process:
  1. [ ] Choose domain and define agent purpose
  2. [ ] Copy agent-template.md to appropriate directory
  3. [ ] Fill YAML frontmatter
  4. [ ] Identify skill integration points
  5. [ ] Document Python tool invocations
  6. [ ] Map workflows to existing skills
  7. [ ] Add integration examples
  8. [ ] Test relative path resolution
  9. [ ] Validate YAML and markdown
  10. [ ] Submit for review
- [ ] Document naming conventions (cs-* prefix, kebab-case)
- [ ] Explain relative path resolution (`../../` patterns)
- [ ] Provide testing checklist
- [ ] Add integration validation steps
- [ ] Include examples from cs-content-creator
- [ ] Document common pitfalls and solutions

## Success Criteria

- [ ] Template covers all required agent sections
- [ ] Creation guide is actionable (developers can create new agents independently)
- [ ] Examples from cs-content-creator are referenced
- [ ] Template passes validation (YAML, markdown linting)
[...]

PR #907, opened 2026-07-08
feat(engineering): vendor SkillOpt-Sleep from microsoft/SkillOpt

alirezarezvani:
## Summary

- Vendors [microsoft/SkillOpt](https://github.com/microsoft/SkillOpt)'s **SkillOpt-Sleep** into `engineering/skillopt-sleep/` — a verbatim, byte-for-byte copy of the `skillopt_sleep` engine (32 files, stdlib-only, zero third-party dependencies) and its Claude Code plugin surface (`skills/`, `hooks/`, `commands/`, `scripts/`).
- Gives a local Claude agent a nightly gated self-improvement cycle: read-only harvest of past Claude Code session transcripts → mine recurring tasks → offline replay on the user's own API budget → consolidate into `CLAUDE.md`/`SKILL.md` edits behind a held-out validation gate → stage for review. **Nothing live changes** until an explicit `/skillopt-sleep adopt` (which backs up first). Default `mock` backend spends no API budget.
- Deliberately did **not** vendor the full `skillopt` *training* package (needs `numpy`/`openai`/`azure-*` + hand-labeled train/val/test benchmark data per task) — it optimizes one narrow, scoreable task at a time, which doesn't fit this repo's broad domain-expertise skills or its no-ML-in-scripts / no-test-framework conventions. `skillopt_sleep` mines its "benchmark" from real repo usage instead, which does fit.
- Attribution preserved in `plugin.json` + `LICENSE` + `README.md` (MIT, © Microsoft Corporation / Yifan Yang), following the same verbatim-vendor pattern already used for `loop-library/`.
- Registered as its own installable marketplace plugin in `.claude-plugin/marketplace.json`.
- Trued up headline counters (355→358 skills, 602→634 Python tools, 109→110 commands, 83→84 plugins) across `README.md`, `CLAUDE.md`, and `marketplace.json` via `scripts/derive_counters.py --check`.

## Checklist

- [x] Skill has `SKILL.md` with valid YAML frontmatter (`name`, `description`)
- [x] Scripts run with `--help` without errors (`python -m skillopt_sleep --help`, `scripts/sleep.sh status`)
- [x] No vendor-locked dependencies without open-source fallback (stdlib-only; `mock` backend has zero API dependency)
- [x] Follows existing directory structure (`engineering/skillopt-sleep/skills/skillopt-sleep/SKILL.md`)

## Type of Change

- [x] New skill
- [ ] Improvement to existing skill
- [ ] Bug fix
- [ ] Documentation
- [ ] Infrastructure / CI

## Testing

- `python3 -m py_compile` on all 32 vendored `.py` files — clean.
- `python3 -m skillopt_sleep --help` — exits 0, prints subcommand usage.
- `scripts/sleep.sh status --project "$(pwd)"` and `scripts/sleep.sh dry-run --project "$(pwd)" --backend mock` — both exit 0, no API spend.
- `hooks/hooks.json` validated as parseable JSON; `hooks/on-session-end.sh` exits 0.
- `python3 scripts/check_plugin_json.py --all` — all 84 `plugin.json` manifests pass.
- `python3 scripts/derive_counters.py --check` — README/CLAUDE.md/marketplace.json counters match the derived tree.

---
_Generated by [Claude Code](https://claude.ai/code/session_01TX374i2YGrjNV4Yi3AmaKS)_


alirezarezvani:
Addressed both automated reviews:

**CI (two rounds of real failures, both fixed):**
- `check_paths.py` flagged `CLAUDE.md/SKILL.md` in the SKILL.md frontmatter as a broken path (its `.../SKILL\.md` regex over-matched prose) → reworded to "CLAUDE.md and SKILL.md" (`91b5aa5`).
- `smoke_scripts.py` tried to run all 29 `skillopt_sleep/*.py` files standalone; the package is designed for `python -m skillopt_sleep` only and `types.py`'s stdlib-name collision breaks direct execution → added all 29 to `smoke_exceptions.txt` with reasons (`16dfdd7`).

**Review findings (`186c0f6`)** — went with "patch all 3, including budget enforcement":
1. `staging.py`: `redact_secrets()` now actually covers `proposed_SKILL.md`/`proposed_CLAUDE.md`, not just diagnostics — closes the gap between the README's "secrets are redacted" claim and what `adopt()`/`--auto-adopt` actually write live.
2. `scheduler.py`: the generated crontab line is now `shlex.quote()`-d (project path, logdir, log, repo root) instead of raw f-string interpolation.
3. `cycle.py`: wired the previously-dead `max_tokens_per_night` config key into the real `run_sleep_cycle()` path — starts a `Budget` right after backend construction, sizes `dream_rollouts` down via the existing (but previously unused) `plan_depth()` heuristic, and the report notes whenever it caps rollouts or the budget is exhausted. This caps rollout depth per task, not a hard mid-call abort — documented as a residual limitation in the plugin README rather than silently claimed as fully solved.

Also dropped a dead hardcoded nvm path in `backend.py` (the generic scan below already covered it) and added a one-line callout in root `CLAUDE.md`'s Anti-Patterns section acknowledging this plugin's non-mock backends as a documented, opt-in exception to "no LLM calls in scripts."

All patches are logged as explicit deviations in `engineering/skillopt-sleep/README.md` so re-vendoring from upstream doesn't silently drop them. Verified: `py_compile` clean, `mock`-backend dry-run still exits 0, a synthetic test confirms the budget capping actually engages under a tight budget and is a no-op under the default budget, and all four repo CI gates (`smoke_scripts`, `check_plugin_json`, `check_paths`, `derive_counters`) pass locally.

---

alirezarezvani:
Addressed the second review pass too (`316982f`):
1. `plugin.json`'s `derivation_note` no longer says "no logic modified" — updated to match `README.md`'s deviation log.
2. `redact_secrets` config key is now actually wired (was declared but unread); disabling it is honored but logs a loud report note rather than failing silently.
3. `scheduler.py`'s `extra` crontab param is now `shlex.quote()`-d too (not exploitable today since it's a hardcoded literal, but closes the class of gap defensively).

Also took the reviewer's non-blocking suggestion: `adopt()` now re-runs `redact_secrets()` on staged content before writing live, rather than a raw `shutil.copy2` — covers a hand-edited staged proposal between `stage` and `adopt`.

All 8 deviations from upstream (2 cosmetic, 6 safety/hardening) are now cross-documented in both `plugin.json` and `README.md` so re-vendoring can't silently drop them. Verified locally: `py_compile` clean, mock dry-run exits 0, synthetic tests for both the extra-quoting and redact-toggle paths, all 4 CI gates pass.

---

alirezarezvani:
Addressed the third review pass (`3244688`), including both HIGH findings:

**HIGH:**
1. `state.py`'s cross-night task archive (`state.json`) persisted raw harvested `intent`/`context_excerpt`/`attempted_solution` text indefinitely, entirely outside the staging dir a user is ever told to review. Now redacted before archiving.
2. `report.md`/`report.json` — the two files a human is told to read *first* — were never redacted, while their siblings (`proposed_SKILL.md`/`proposed_CLAUDE.md`/`diagnostics.json`) were fixed in rounds 1-2. Now redacted too.

**Medium/low:**
3. `replay_mode: "fresh"` is genuinely unimplemented (no worktree replay exists anywhere in the engine) — real worktree isolation was judged too invasive to build into a vendored copy, so instead the report now warns loudly whenever it's set to anything but `"mock"` rather than silently implying isolation that isn't happening.
4. Removed a hardcoded internal Azure OpenAI backend (5 internal-looking endpoint hostnames + a Managed Identity client ID, commented as sourced from "the intern's avail_api.md") — it was already unreachable from this plugin's documented `--backend` choices and needs deps we don't vendor, so I dropped it rather than carry forward what reads like leaked Microsoft-internal infra.
5. Tool names reachable via `--tasks-file` (used as both a shim filename and unescaped shell text) are now validated against a safe-identifier allowlist across all three CLI backends.
6. Fixed the dead `docs/superpowers/specs/...` cross-reference in `SKILL.md` (points to the real upstream guide URL now).
7. Added the missing dated `v2.11.2` section to `CLAUDE.md` matching the repo's per-release convention.

All 13 deviations (2 cosmetic, 11 safety/hardening across three review rounds) are cross-documented in `plugin.json` and `README.md`. Verified with a synthetic test using a real API-key-shaped secret seeded into a task's `intent` — confirmed it appeared in none of `state.json`'s archive, `report.md`, or `report.json` after the fix (all three showed positive before it). All 4 CI gates pass locally.

---

alirezarezvani:
Addressed the fourth review pass (`e729f15`):

1. **Deviation count drift** — you were right, `CLAUDE.md` said "8 deviations" while `README.md`/`plugin.json` said 13, and one item (the SKILL.md dead cross-reference) was fixed in round 3's commit but never actually added to README's numbered list. Added it as deviation #14, and made README.md's numbered list the explicit single source of truth in `plugin.json`'s note (so a future disagreement resolves deterministically instead of drifting again).
2. **`schedule` confirm-first inconsistency** — the command doc's action table and its "Safety reminders" section told two different stories about the same action. `scheduler.schedule()` does write straight to the user's real crontab with no confirmation gate. Fixed: "Steps to follow" now has an explicit step 1 requiring the agent to confirm with the user before running `schedule` (naming what will be scheduled), and "Safety reminders" no longer contradicts that.
3. **No restrictive file permissions** — `state.json` and the staging dir's proposal/report/diagnostics files carry real harvested session content in plaintext at default (often world-readable) umask permissions. Fixed: `state.py`/`staging.py` now `chmod 0700` every directory they create and `0600` every file they write. Verified with a synthetic run that the actual on-disk modes are correct. Left live `CLAUDE.md`/`SKILL.md` untouched since those are the user's own, often-committed files.

All three documents (README's numbered list, `plugin.json`, `CLAUDE.md`) now agree on **16 deviations (3 cosmetic, 13 safety/hardening)** — cross-checked with a grep, not just eyeballed. All 4 CI gates pass locally.

---

alirezarezvani:
Thanks — nice to hear the earlier rounds are holding up under independent re-verification. Applied the one cosmetic item anyway since it was trivial (`e64c1cd`): `staging.py`'s `sk-...` redaction pattern also matches Anthropic/other vendor keys, so `[REDACTED_OPENAI_KEY]` was a misleading label even though redaction itself worked correctly either way. Relabeled to `[REDACTED_API_KEY]`, added as deviation #17, and reconciled the count to 17 (4 cosmetic, 13 safety/hardening) across all three documents.

---

alirezarezvani:
Addressed the sixth review pass (`19fd75a`):

1. **Stale fallback paths in `sleep.sh`/`run-sleep.sh`** — confirmed: both scripts' header comments and `SKILLOPT_SLEEP_REPO`/`CLAUDE_PLUGIN_ROOT` fallback branches still assumed upstream's `<repo>/plugins/claude-code/` three-level layout instead of this vendored copy's actual `scripts/` + `skillopt_sleep/` sibling layout. The primary co-located/repo-relative resolution paths always succeed regardless, so this was unreachable in normal operation — but I fixed it properly rather than leaving dead-but-wrong code: `sleep.sh` now checks `$SKILLOPT_SLEEP_REPO/scripts/run-sleep.sh`, `run-sleep.sh` now checks `$CLAUDE_PLUGIN_ROOT/skillopt_sleep` directly (this repo's real layout) ahead of the upstream two-levels-up check. I verified both previously-broken branches actually resolve now by copying each script to an isolated scratch dir and running it standalone with only the relevant env var set (so the co-located fallback couldn't mask a still-broken path).
2. **"711 templates" vs "731 reference guides"** — correctly identified as pre-existing and out of scope for this PR; left untouched as you suggested.
3. **Risk-category judgment call** — noted. This plugin's real-world-mutating actions (crontab install, `adopt`) were a deliberate, discussed part of what got vendored here from the start (not a scope-creep add-on), and both are now gated behind explicit human confirmation in the command doc rather than run unattended by default.

18 deviations (5 cosmetic, 13 safety/hardening) across six review rounds, counts still cross-checked and agreeing across all three documents.

---

alirezarezvani:
Trimmed `plugin.json`'s description as suggested (`73c2570`) — 930 → 705 chars, moved the detailed derivation story fully into `attribution.derivation_note` (which already carried it) rather than duplicating it in both places.

The other two minor items (`harvest.py`'s timestamp parsing already fails safe; the `extra`-param assumption is already documented in the code comment you quoted) don't request a change, so left as-is.

**The "no LLM calls in scripts" exception** — flagging this explicitly for @alirezarezvani rather than deciding it myself, since the reviewer is right that it's a first-of-its-kind carve-out worth a conscious yes: `backend.py`'s `claude`/`codex` backends do shell out for real replay when a non-`mock` backend is explicitly chosen. It's disclosed (not hidden) in both `plugin.json` and `CLAUDE.md`'s Anti-Patterns section, scoped to opt-in only, and arguably unavoidable given the feature's whole premise — but whether that's precedent other future skills can point to, or a one-off exception for this specific self-improvement-loop category, is a repo-policy call rather than a code-correctness one. Surfacing it here rather than merging past it silently.

---

alirezarezvani:
Confirmed and fixed the HIGH finding (`4d68c54`) — good catch, this was real. `write_staging()`'s redaction operates on `report.to_dict()`, a copy used only for the on-disk JSON; it never touches `report.edits`/`report.rejected_edits` themselves, so `__main__.py::cmd_run()` — reading the same in-memory `Report` object — was printing unredacted `EditRecord.content` to the console and serializing it unredacted for `--json`. Exactly as you traced: `scheduler.py`'s cron entry redirects that stdout straight into `cron.log`, which (unlike `state.json`/staged files) also had no `chmod` protection.

Fixed both halves:
- `_report_payload()` and `cmd_run()`'s plain-text edit printing now route through `redact_secrets()`.
- I also extended the same fix to `cmd_harvest()` (same bug shape — prints raw mined `TaskRecord.intent` for `--json`/`--output`/plain-text), since redaction only strips secret-shaped substrings and doesn't reduce what a human needs to review there.
- `scheduler.py`'s generated cron line now `chmod 700`s the log dir and `chmod 600`s `cron.log` itself before each run.

Verified with a seeded secret that it's actually gone from `cmd_run`'s `--json`/plain-text output and `cmd_harvest`'s payload, and by executing the real generated cron line end-to-end (not just inspecting the string) that the log dir/file land at `0700`/`0600` on disk.

On the low-severity `managed_skill_path()` containment nit — agree with your own assessment that it's non-blocking (deliberate config/CLI input only, not attacker-controlled harvested content), so left it as a known limitation rather than adding a policy call about legitimate out-of-project targets that wasn't asked for.

19 deviations (5 cosmetic, 14 safety/hardening) across eight review rounds, counts reconciled across all three documents.

---

alirezarezvani:
Confirmed and fixed (`a227b06`) — good catch, and nice cross-reference to `harvest.py`'s `_project_matches()` getting the same class of problem right elsewhere. Reproduced the bug standalone first (two sibling-prefix projects, scheduling/unscheduling one silently dropped the other's line), then fixed it with an anchored `ln.rstrip().endswith(marker)` check at both call sites, then verified via a full `schedule()`/`unschedule()` round-trip through the real public API (swapped `crontab -l`/`crontab -` for an in-memory fake) that `app-v2`'s line survives unscheduling `app`.

Also quoted `install-cron.sh`'s `--backend` value for consistency with `_runner_cmd`'s discipline, as you suggested.

21 deviations (6 cosmetic, 15 safety/hardening) across nine review rounds, counts still reconciled across all three documents.

---

alirezarezvani:
Fixed both (`7eb198f`) rather than deferring — thanks for the option, but keeping everything closed out now:

1. **`schedule` had no CLI-layer confirmation gate** — correct, the round-4 safeguard only lived in the agent-facing command doc. `schedule` now requires `--yes` (interactive terminals get a `[y/N]` prompt instead; non-interactive without `--yes` refuses outright with exit 2). Updated `commands/skillopt-sleep.md` so the driving agent passes `--yes` *after* confirming with the user in chat — that confirmation is what `--yes` records, not a redundant prompt that would hang forever with no TTY inside a non-interactive Bash tool call.
2. **mkdir-then-chmod non-atomicity** — added `mode=0o700` to the `os.makedirs()` calls creating the state/staging/backup dirs, on top of (not instead of) the existing post-creation `chmod` calls (which still matter for intermediate parents and pre-existing dirs `mode=` doesn't reach). I judged the equivalent per-file race not worth the larger rewrite (every `open()` call site would need `os.open()` with an explicit mode) for a finding you'd already flagged as requiring precise-timing local access — documented as a narrower, still-open residual gap rather than silently claimed as fully closed.

Verified both: non-interactive `schedule` without `--yes` refuses with exit 2 / with `--yes` proceeds normally; a synthetic run confirms permissions are unchanged (still `0700`/`0600`) after the `mode=` change.

23 deviations (6 cosmetic, 17 safety/hardening) across ten review rounds, counts still reconciled across all three documents.

---

alirezarezvani:
Thanks for the thorough independent re-verification — glad the ten rounds of fixes hold up under a fresh read rather than just trusting the commit messages.

On the two minor notes, I'm leaving both as-is rather than adding more churn, since neither is actually asking for a change and I think the current behavior is the right call:

1. **`adopt` vs `schedule` confirmation asymmetry** — intentional, not an oversight. `schedule` installs an *unattended, indefinite, system-wide* crontab entry that can run for months with no further human involvement, which is why it needs its own gate. `adopt` is a *single, explicit, project-local* action, always preceded (per the documented workflow — see `commands/skillopt-sleep.md` step 4) by showing the user `report.md`'s exact proposed edits and only ever being offered after that review, and it always backs up before writing. Typing `adopt` *is* the confirmation; gating it again would add friction to the plugin's designed "review via report.md, then apply" flow without closing any real gap — there's no path where `adopt` fires on content the user hasn't already seen.
2. **Broad secret-redaction regex** — the false-positive-favoring tradeoff (over-redact rather than under-redact) is deliberate and correct for this use case; not something I want to narrow without a concrete case where it's actually causing a problem.

No code changes this round. 23 deviations across ten review rounds still stand, all cross-checked and consistent.

---

PR #939, opened 2026-08-03
fix(agents): resolve skills: preloading, namespace colliding skill names

benrfairless:
## Summary

> **Stacked on #938.** Review the top 3 commits (`4415e0e0`, `43138ac8`, `09f372dd`). The third is pure regenerated output.

**1. `skills:` on agents is now a live spec field with values that never resolve.**

When these agents were written, `skills:` was inert repo documentation. The current [subagent spec](https://code.claude.com/docs/en/sub-agents#write-subagent-files) interprets it: listed skills are **preloaded**, with their full body injected into the subagent's context at startup, resolved by skill *name*. Every value in the repo was a repo-relative *path* instead (`skills: c-level-advisor/skills/cfo-advisor`), a handful were bare domain names, and two named skills that do not exist. All silently resolved to nothing.

All **112 references across 81 agents** now use the `plugin:skill` form. Domain-level values point at that domain's index skill; the two phantom entries (`healthcheck`, `cost-estimator`) are dropped; dual-published skills resolve to the bundle, matching the path the author originally wrote.

⚠️ **This turns preloading on** — the behaviour the field was always documenting. Measured cost is **~2,500 tokens per agent spawn**. Happy to invert this commit to deletion instead if you would rather opt in deliberately; the mechanical work is the same either way.

**2. `context: fork` on 11 agents.** It is a *skill* field, absent from the agent frontmatter table, silently ignored there. Removed.

**3. Plugin-name collision.** `engineering/handoff` and `productivity/handoff` both declared `name: handoff`, so their skills namespaced identically and collided whenever both were installed. Aligned with their existing marketplace entries (`handoff-engineering` / `handoff-productivity`) — which is also what `enabledPlugins` already keys on, so this is a fix rather than a break.

**4. Six colliding bare-verb skill names.** `init`, `run` and `status` were each declared by two plugins, and `/init` and `/run` are also Claude Code bundled skills. Renamed to the prefix each plugin already uses in its own docs:

| plugin | before | after |
|---|---|---|
| `agenthub` | `init` / `run` / `status` | `hub-init` / `hub-run` / `hub-status` |
| `autoresearch-agent` | `run` / `status` | `ar-run` / `ar-status` |
| `playwright-pro` | `init` | `pw-init` |

Also fixes `playwright-pro`'s `pw` skill, which declared `name: playwright-pro` from a directory named `pw`.

**Finding worth a separate decision:** both plugins document their commands as `/hub:init` and `/ar:run`, but the namespace comes from the *plugin* name — so the real commands are `/agenthub:hub-init` and `/autoresearch-agent:ar-run`. That mismatch predates this PR. Fixing it properly means renaming the plugins to `hub` and `ar`, which would also make the skill renames above unnecessary. Left alone here.

**5. Mirror resync (separate commit).** The renames dangled 20 symlinks across `.codex`, `.gemini`, `.vibe` and `.hermes`. The sync scripts add links but never prune, so the dead ones are removed explicitly and all four regenerated. `sync-vibe-skills.py` and `sync-hermes-skills.py` default to `$HOME`, so they were pointed at the in-repo mirrors with `--target`. Only the Codex sync runs in CI, which is why the other three had already drifted independently.

## Checklist

- [x] Skill has `SKILL.md` with valid YAML frontmatter
- [x] Scripts run with `--help` without errors
- [x] No vendor-locked dependencies without open-source fallback

## Type of Change

- [ ] New skill
- [x] Improvement to existing skill
- [x] Bug fix
- [ ] Documentation
- [x] Infrastructure / CI

## Testing

```
check_frontmatter.py --all       0   593 files, 0 errors, 6 warnings (was 17)
check_paths.py --all             0
check_model_freshness.py --all   0
check_dual_publish.py            0
smoke_scripts.py                 0
derive_counters.py --check       0
check_plugin_json.py --all       0
```

Every one of the 112 `skills:` values was validated against the set of real `plugin:skill` pairs built from the tree — 0 unresolvable. Broken symlinks across all four mirrors: 20 before, 0 after.

`check_paths.py` earned its keep here: a global string replace corrupted body paths in two product agents mid-work, the gate failed, and the replace was rescoped to the frontmatter line.

---

*Authored with Claude Code (`claude-opus-5`). All gates above were run locally before submission.*


alirezarezvani:
This PR went through the repo's full open-PR audit (`audit/pr-stream-2026-08/` on `dev`) and came out **MERGE-WITH-CHANGES** — the mechanical work was independently re-verified (all 109 agent `plugin:skill` references resolve against the real plugin.json + SKILL.md name set, zero unresolvable; handoff plugin collision genuinely fixed; broken mirror symlinks 3→0). #936–#938 from your stack are now merged into `dev` — thank you, every claim in them held up under re-execution.

Decisions and remaining items to get this one in:

1. **Preloading decision (recorded): keep the `plugin:skill` form ON.** The ~2.5k tokens/agent-spawn cost is accepted; resolvable references beat deleting the field. No need for the deletion variant.
2. **Rebase onto current `dev`** — it has moved substantially since your base (book-to-skill, memory-engineering, your own #936–#938, plus a batch of manifest/naming fixes from #972/#973 that renamed several built-in-shadowing skills). Please resolve the `.hermes/skills/claude-skills/skills-index.json` conflict by re-running the four sync scripts rather than hand-merging, and re-check your rename set against #973's renames so they don't collide.
3. **Cover agents added since your base** — e.g. `engineering/book-to-skill/agents/cs-book-to-skill.md` still carries the old path form and would re-land the bug this PR fixes.
4. **Multi-skill encoding:** 9 agents got multi-skill values flattened to one comma-joined string (e.g. `agents/personas/content-strategist.md`). Please re-encode those as YAML lists — safe under either parser reading.
5. A short release note for the renamed invocations (`pw`, `hub-*`, `ar-*`, `handoff-*`) in the PR body would help downstream users.

Happy to re-review as soon as it's rebased.

---

alirezarezvani:
Heads-up before your next rebase: part of item 4 landed on `dev` independently via PR #973 (merged 2026-08-21), which renamed the built-in-shadowing skills as part of closing issue #885 — `agenthub` init/status → `hub-init`/`hub-status`, `autoresearch-agent` status/resume → `ar-status`/`ar-resume`, `playwright-pro` init/review → `pw-init`/`pw-review`. Good news: the conventions match yours exactly, so on rebase those hunks should collapse to no-ops rather than conflicts of substance. Still uniquely yours and wanted: the `run` renames (`hub-run`/`ar-run` — `run` isn't a Claude Code built-in so the new `scripts/check_skill_names.py` CI gate from #973 doesn't cover cross-plugin collisions like it), the `pw` skill's name/directory mismatch, the `handoff` plugin-name collision, and the whole `skills:` preloading fix (items 1–3). #973 also regenerated all four mirror trees for its renames, so item 5 will need re-running on the rebased result.

---

alirezarezvani:
Reviewed the top 3 commits and independently verified the premises against current `dev`. **The diagnosis is correct on all four counts** — but I measured the preloading cost and got a materially different number than the one in the description, which I think should change how the open question gets decided.

## Premises verified against `dev`

| Claim | Verified |
|---|---|
| `skills:` values are not resolvable names | ✅ current values are a mix of bare names (`ceo-advisor`) **and** repo paths (`marketing-skill/skills/marketing-demand-acquisition`) — neither is the `plugin:skill` form the spec resolves |
| `healthcheck` / `cost-estimator` are phantoms | ✅ **0** `SKILL.md` in the tree declares either name |
| `handoff` plugin-name collision | ✅ both `engineering/handoff` and `productivity/handoff` declare `name: handoff` |
| Bare-verb skill collisions | ✅ `name: "run"` is declared twice |

53 files under the domains I sampled carry a `skills:` block, consistent with your 81-agent count across the full tree.

## ⚠️ The preload cost looks higher than ~2,500 tokens/agent

I built a name→`SKILL.md` index and summed the bodies each agent would inject. **Methodology caveat up front:** I matched on bare skill name against current `dev`, so only **6** agents resolved (the subset whose existing values happen to match a real skill name), and I used `chars/4` as a token proxy. This is a *sample of the current state*, not a measurement of your post-PR tree — you have the accurate denominator, I don't.

With that caveat, the resolved sample:

```
mean          ~8,673 tokens/agent spawn
heaviest:
  ~19,064 tok  content-strategist.md   (9 skills)
  ~16,299 tok  product-manager.md      (7 skills)
  ~ 7,198 tok  cs-demand-gen-specialist.md (3 skills)
  ~ 5,286 tok  devops-engineer.md      (5 skills)
  ~ 2,152 tok  cs-product-analyst.md   (2 skills)
```

The distribution is what concerns me more than the mean: it's driven by **skill count per agent**, so the agents that list 7–9 skills pay 6–8× the quoted figure. If `~2,500` was computed as a mean across all 81 agents (many listing a single small skill), both numbers can be right — but the tail is what users actually feel, and a 19K-token fixed cost on every `content-strategist` spawn is a different proposition from 2.5K.

**Could you share how the 2,500 figure was derived?** If it's a per-*skill* average rather than per-*agent*, the description's "per agent spawn" framing understates it and the maintainer is deciding against the wrong number.

## On the preload-vs-delete question

Given the above, my read is that **deletion is the safer default** and the one I'd recommend, for three reasons:

1. **The field has never worked.** Nothing regresses by removing it — no user has ever received a preloaded skill from these agents. Turning it on is a *new* behavior with a new cost, not a restoration.
2. **The cost is unbounded by design.** It scales with however many skills an author lists, and nothing in the repo's conventions caps that. Today's worst case is ~19K tokens; nothing stops a future agent from listing 15 skills.
3. **Agents can already reach skills on demand.** Preloading buys guaranteed availability at the cost of paying for every skill on every spawn, whether used or not — the opposite of the progressive-disclosure principle the repo applies elsewhere (`write-a-skill`'s references/ pattern exists precisely to avoid resident bulk).

If you do want preloading on, I'd suggest gating it: keep `skills:` only on agents listing ≤2 skills, and drop it from the 7–9 skill agents where the cost concentrates.

**Everything else in this PR is unambiguously good** and I'd take it regardless of which way that decision goes — the collision fixes (items 2–4) are real bugs with no downside, and the mirror resync with explicit pruning addresses a genuine gap (the sync scripts add but never prune).

## Separate note on the `/hub:init` vs `/agenthub:hub-init` finding
Agreed this is out of scope here, and agreed that renaming the plugins to `hub`/`ar` is the better end state since it would make the skill renames unnecessary. Worth its own issue so it isn't lost — happy to file one.

---
*Verification run locally against `dev` @ `7dd5eec`. The token figures are a sample with the stated methodology limits, not an authoritative count.*

---

alirezarezvani:
Maintainer decision on this PR — the work here is valued, and we'd like to land the surviving parts, but the branch needs a rebase and a scope cut:

**Decision on your ⚠️ preload question: invert to deletion.** Please **delete the dead `skills:` values** rather than converting them to resolving `plugin:skill` form. This library's agents are meant as lightweight routers, and silently adding ~2,500 tokens of preloaded skill body to every one of 81 agent spawns is a cost nobody opted into. Thank you for flagging it as a choice instead of just switching it on — that was the right call.

**Superseded (please drop from the rebase):** item 4's skill renames landed via #973 during the issue-triage sweep (issue #885): `hub-init`/`hub-status`, `ar-status`/`ar-resume`, `pw-init`/`pw-review`, `memory-status`, plus a new blocking CI gate (`scripts/check_skill_names.py`) that guards the reserved-word set going forward. Your `hub-run`/`ar-run` renames beyond that set may still be worth carrying if they collide with bundled skills — check against the gate's RESERVED list on current `dev`.

**Still wanted, scoped rebase onto current `dev`:**
1. Delete the ~112 dead `skills:` path values across the agents (per the decision above).
2. Remove the ignored `context: fork` from the 11 agent files.
3. The `handoff` plugin-name collision fix (`handoff-engineering` / `handoff-productivity`) — verified still live on `dev`.
4. The mirror-tree prune (dangling symlinks; the sync scripts add but never prune). Note `dev` has moved ~3 weeks since your base — including the v2.12.0 release train — so regenerate rather than replay the old commit.

Also for context: #940 was closed, but its two verified findings (stale `.codex-plugin/plugin.json`, ungated `mkdocs.yml`) were adopted in #989 with credit.

If you'd rather not carry the rebase, say so and we'll cherry-pick the surviving work with attribution. Thanks again — the `skills:`-field investigation in particular was a genuinely valuable find.

---

PR #941, opened 2026-08-05
feat(engineering): add book-to-skill — compile documents into knowledge-base skills

alirezarezvani:
## Summary

Adds `engineering/book-to-skill` — our own version of [virgiliojr94/book-to-skill](https://github.com/virgiliojr94/book-to-skill) (MIT), adapted to this repo's conventions and wired into the rest of the library.

Compiles a book, documentation folder, or spec collection (PDF, EPUB, DOCX, HTML, Markdown, RST, AsciiDoc, RTF, MOBI/AZW) into an agent skill: a resident master `SKILL.md` (core frameworks + chapter index + topic index, capped at 4k tokens) plus on-demand `chapters/chNN-*.md`, `glossary.md`, `patterns.md`, and a decision `cheatsheet.md`. The agent reads the core, then one chapter — never the whole source again.

**The "ingest into other plugins and skills" part** is `Step 11` / `/cs:book-to-plugin`, which has no upstream counterpart. Upstream stops at a bare folder in `~/.claude/skills/` that this library cannot route to. `skill_plugin_emitter.py` wraps a compiled skill as a full plugin package — manifest + `cs-<slug>` agent + `/cs:<slug>` command + README — and prints the marketplace entry to register. So a user of this repo can point the skill at their own book or internal docs and get back an installable plugin.

### What ships

| | |
|---|---|
| Tools (4 CLIs + a stdlib-only library package) | `extract_document.py`, `book_skill_validator.py`, `token_budget_estimator.py`, `skill_plugin_emitter.py` |
| References (5, 7–8 sources each) | conversion workflow · knowledge-extraction canon · progressive-disclosure budgets · document-extraction pipeline · rights and provenance |
| Assets (3) | chapter / master-skill / cheatsheet templates |
| Agent + commands | `cs-book-to-skill`, `/cs:book-to-skill`, `/cs:book-to-plugin` |

The extraction library (`scripts/book_to_skill/`, 7 per-format parsers) is vendored close to verbatim — it keeps upstream's format chains, chapter detection across Latin/Roman/Chinese/Thai/Korean heading styles, invisible-Unicode (Trojan Source) sanitization, and the DOCX entity-expansion guard. That code is good and re-deriving it would only make it worse.

### Deviations from upstream

**18 numbered items**, authoritative list in `engineering/book-to-skill/README.md`. The ones worth reading:

- **No implicit installs (5).** `--install-missing` defaults to `report` — prints the pip command, uses the stdlib fallback. Upstream prompts on a TTY and runs `pip install` into the caller's environment. Reading a file should not install packages.
- **Rights gate (6).** Emitting a *shareable* package refuses without `--rights` from `public-domain|open-license|internal-docs|author-permission`. `fair-use` is deliberately excluded — it's a defence, not a licence, and not a script's call.
- **Validator merged and extended (10).** Upstream's two validators become one four-family gate, adding **budget** (token caps) and **index** (dead chapter links, unindexed chapter files, dangling topic refs). Index integrity is the failure that silently breaks navigation while the skill still *looks* complete; upstream had no check for it.
- **Folded YAML scalars now parse (11)**, so a wrapped description no longer under-reports its length past the 1024-char cap — the bug only bit descriptions long enough to hit the cap.
- **Worth-converting verdict (12).** `discovery_tax.py` → `token_budget_estimator.py`: `tiktoken` path dropped for one deterministic estimator, post-flight budget audit added, plus an explicit verdict that says "just read it" when the source is under ~3× the compiled skill. Upstream reports only savings ratios, which read as advocacy on a source too small to convert.
- **Private per-invocation workdir (15).** Upstream's fixed `<tempdir>/book_skill_work` is CWE-377/CWE-59 on a shared host. Now `mkdtemp` (0700, unpredictable) with 0600 artifacts; explicit `--workdir` is symlink-refused and mode-restricted. Also fixed a real bug: `parsers/calibre.py` read a module-level path constant and so ignored `--workdir` entirely.
- **Zip-of-XML hardening generalized (17).** Upstream's DTD/entity guard covered DOCX only; EPUB's `ebooklib` path had none. Now shared in `zip_safety.py` for both, and every archive read checks declared size and compression ratio before decompressing — a 200 MB zip bomb is refused at ~14 MB peak RSS.
- **Packaging refuses symlinked trees (18).** `shutil.copytree` defaults to following links, which would bake a link target's content into a distributable package. The whole tree is now walked and refused, before the validation branch so `--skip-validation` cannot bypass it.
- Two `PRIV-ESC` criticals fixed: upstream's install hints contained a literal `sudo apt install`. They now name the package manager without telling anyone to escalate.

### Cross-links

- `engineering/write-a-skill` gained a section pointing here: **author first, compile second.** That skill authors from expertise in your head; this one compiles from a document on disk.
- Engineering harness manifest regenerated — picked up `book-to-skill` plus three skills that had drifted out (`minimalist`, `skillopt-sleep`, `strict-api`): `skill_count` 81 → 85.
- Registered in `marketplace.json` and in the `.codex` / `.gemini` / `.hermes` / `.vibe` indexes.

## Checklist

- [x] Scripts run with `--help` without errors — all four CLIs also support `--sample` and `--output json`
- [x] No vendor-locked dependencies without open-source fallback — every format degrades to a stdlib parser; MOBI/AZW is the one exception and needs Calibre, which is stated everywhere it matters

## Type of Change

- [x] New skill
- [x] Improvement to existing skill (`write-a-skill` cross-link; harness manifest drift trued up)

## Testing

**Tool contract** — all four CLIs: `--help` exit 0, `--sample` exit 0, `--sample --output json` parses as JSON.

**End-to-end**, on a generated 3-chapter source:

1. `extract_document.py` → 3 chapters detected, ToC found, artifacts written to a private workdir
2. `token_budget_estimator.py --full-text` → correctly returned **"probably not worth converting"** for a source below the ~3× threshold
3. Built a compiled skill matching the template → `book_skill_validator.py` **passed all four families, exit 0**
4. `token_budget_estimator.py --skill-dir` → per-file budget table, all under cap
5. `skill_plugin_emitter.py` → emitted a 9-file package; **both** the shareable and local manifests pass `scripts/check_plugin_json.py`; the copied skill still validates in place

**Adversarial checks** (each run against the actual attack, not just the patch):

| Attack | Result |
|---|---|
| Planted `full_text.txt -> victim` symlink in the workdir | `REFUSED`; victim file untouched |
| Symlinked `--workdir` | `REFUSED` |
| Pre-existing `777` workdir | tightened to `700`; artifacts `600` |
| Symlink in `assets/` pointing at a secret, then packaged | `REFUSED` with and without `--skip-validation`; secret never copied |
| EPUB whose OPF declares an entity | `REFUSED`; clean EPUB still extracts |
| 200 MB / 1029× DOCX zip bomb | `REFUSED` at ~14 MB peak RSS |

**Gates:**
- `derive_counters.py --check` → passes (skills 362→363, tools 644→663, refs 741→746, agents 102→103, commands 116→118, plugins 88→89)
- `check_plugin_json.py --all` → 89 OK, 0 FAIL
- `check_paths.py --all`, `check_dual_publish.py`, `smoke_scripts.py`, `compileall` → all clean

**Security audit** (`skill-security-auditor` on `skills/book-to-skill/`): **0 critical, 4 high**, all reviewed and documented in the plugin README — 1× `FS-ABUSE` (`shutil.rmtree` on the `--force` path, guarded by `_assert_replaceable()`) and 3× `DEPS-RUNTIME` (one docstring sentence and two `print()` calls that *show* an install command; none install anything).

## Known gaps

- **`write-a-skill` 6-item checklist: 4/6.** `SKILL.md` is 221 lines (was 549 before the full workflow moved to `references/conversion_workflow.md`) against the 100-line rule, and the terminology check flags `skill`/`tool` — both concepts this skill genuinely needs. `write-a-skill` itself fails the same two items, so this matches the repo's real baseline rather than clearing the gate.
- **`.hermes` / `.vibe` indexes carry pre-existing drift** (306 and 339 vs 363 skills). Their sync scripts write to `$HOME`, not the repo, so the committed trees are snapshots from the maintainer's machine. I added this skill's symlink + index entry by hand with a minimal diff and left the older drift alone rather than mixing an unrelated 500-line regeneration into this PR.

https://claude.ai/code/session_017zu9Gmm9S78c2t3kDLnpPX


alirezarezvani:
Thanks — the temp-workdir finding was correct, and worse than described. Pushed in `b5031e9`. All three points addressed.

## 1. Temp workdir (the blocking one) — fixed

You were right about the shape and right that this PR didn't get the `skillopt-sleep` treatment. Took your first suggestion (`mkdtemp`) rather than the minimum, because the fixed name also caused silent collisions between concurrent runs, which mode bits don't solve:

- **Default** is now `tempfile.mkdtemp(prefix="book_skill_work_")` — unpredictable, `0700` by construction, never shared with a concurrent run. The path is printed on completion and carried in `metadata.json`'s `output_text`, so nothing downstream needs a hardcoded path.
- **Artifacts** are written `0600`, and every write refuses a symlink at the target — so even a workdir the user pins can't be used as a write-through.
- **An explicit `--workdir` / `BOOK_SKILL_WORKDIR`** is still honoured, but symlink-refused, created `0700`, and `chmod`-tightened if it already exists (mkdir's `mode=` only applies on creation and is masked by umask, so an existing loose directory needed the explicit tighten).

**One thing you didn't name that turned out to be a live bug.** `parsers/calibre.py` didn't just use the same fixed path — it read the *module-level* `OUTPUT_DIR` constant, so its `ebook-convert` scratch file ignored `--workdir` entirely and escaped whatever directory the caller asked for. That's a bug my own `--workdir` flag introduced when I added it to the vendored library. It now uses its own `TemporaryDirectory`, which removes both the fixed path and the divergence.

`OUTPUT_DIR` / `OUTPUT_TEXT` / `OUTPUT_META` are gone as module-level constants; `config.py` exposes filenames plus a resolver, so there's no fixed path left for anything to import.

Verified, including your exact attack:

| Check | Result |
|---|---|
| Default workdir mode / artifact mode | `700` / `600` |
| Per-invocation name | `/tmp/book_skill_work_2tbi_syc` |
| Two consecutive runs | distinct directories |
| `--workdir` on a pre-existing `777` dir | tightened to `700` |
| Symlinked `--workdir` | `REFUSED`, exit 2 |
| Planted `full_text.txt -> victim` symlink | `REFUSED`; victim file untouched |

## 2. `smoke_exceptions.txt` comment — corrected

Fair catch, the rationale claimed more than it delivered. Rewritten to say what the list actually is: modules the G8 probe *trips on* (those with absolute `book_to_skill.*` imports), not modules that aren't CLIs. The comment now names the six that pass only because they have no argv handling — `config.py`, `exceptions.py`, `sanitize.py`, `parsers/__init__.py`, `parsers/pdf.py`, `parsers/text.py` — and states that the not-a-CLI set is the whole package.

## 3. Duplicated `BUDGETS` — deduplicated

Took this one rather than leaving it as a nit: two tools that *gate* on the same numbers is exactly where silent drift costs something. Now `SKILL_FILE_BUDGETS` and `CHAPTER_TOKEN_CEILING` in `book_to_skill/config.py`, imported by both. Asserted at runtime that the two modules resolve to the same object.

---

Recorded as deviations 15 and 16 in the plugin README (count synced in `plugin.json`, `CLAUDE.md`, `CHANGELOG.md`). Docs that referenced the old fixed path now read it from the tool's output.

Re-verified after the change: all four tools pass `--help` / `--sample` / `--sample --output json`; the end-to-end pipeline (extract → verdict → validate → emit → manifest check) runs clean; and every blocking gate is green — `compileall`, `check_paths --all`, `check_dual_publish`, `smoke_scripts` (0 failed), `derive_counters --check`, `check_plugin_json --all` (0 FAIL). Security auditor unchanged at 0 critical / 4 high, all four documented in the plugin README.

---

alirezarezvani:
All four addressed in `9ed709a`. Both mediums were real; I verified each against the actual attack rather than just applying the patch.

## 1. [Medium] `copytree` dereferencing symlinks the validator never checked — fixed

Correct, and the `--skip-validation` observation is the part that made it worth more than a flag change. Went with **refuse** rather than `symlinks=True`: preserving the link would ship a package whose contents depend on the emitting machine's filesystem, which is its own problem for something distributable.

`_assert_no_symlinks()` walks the whole tree and refuses, and it runs **before** the validation branch — so `--skip-validation` waives content findings but never the guarantee that packaging copies only what's actually in the tree. `copytree` also now passes `symlinks=True` as belt-and-braces, so a future edit loosening the walk can't silently reintroduce dereferencing.

```
$ ln -s /tmp/secret.txt skill/assets/innocuous.md
$ skill_plugin_emitter.py --skill-dir skill --dest out
REFUSED: source skill contains 1 symbolic link(s); packaging would copy the target's
content into a distributable package:
  assets/innocuous.md -> /tmp/secret.txt
$ skill_plugin_emitter.py --skill-dir skill --dest out --skip-validation
REFUSED: ... (same)
```

Secret content absent from both attempted packages; a clean tree still emits normally.

## 2. [Medium] DOCX XXE guard not extended to EPUB's `ebooklib` path — fixed

Also correct, and the framing was the useful part: `ebooklib` is a package this skill *recommends installing*, so the better-quality path was the less-guarded one. Upstream hardened DOCX and stopped.

The guard moved to a new `book_to_skill/zip_safety.py` and now runs for both formats — `validate_zip_xml_safety(path, label)` scans every XML-ish member (`.xml`, `.rels`, `.opf`, `.ncx`, `.xhtml`, `.html`) across the same five encodings before any parser touches the archive.

```
$ extract_document.py entity.epub
WARNING: skipping entity.epub: security validation failed: XML member 'c.opf' in the
EPUB archive contains a forbidden DTD or entity declaration

$ extract_document.py clean.epub
   Chapters: 1 detected overall
```

## 3. [Low] Stale deviation count in the PR body — fixed

You were right that it was stale, and it had drifted further since: the in-repo list is now **18**. PR description updated (and it was 16 at the time you looked, not 12 — the body was two rounds behind, not one).

## 4. [Low] No size cap before decompressing zip members — fixed

Took this one rather than deferring, since I was already in `zip_safety.py`. Every archive read now goes through `safe_read()`, which checks the central directory's declared `file_size` and the `file_size/compress_size` ratio **before** decompressing, then charges the actual bytes against a per-archive budget — so a lying directory entry can't get past it either. Caps: 64 MB/member, 512 MB/archive, 200× ratio.

```
$ # 203,842 bytes stored -> 209,715,200 uncompressed (1029x)
$ extract_document.py bomb.docx
WARNING: skipping bomb.docx: archive member 'word/document.xml' declares 209,715,200
bytes uncompressed; the per-member limit is 67,108,864 bytes
peak child RSS: 14 MB
```

Refused at ~14 MB instead of materializing 200 MB.

---

Recorded as deviations 17 and 18 (count synced across `plugin.json`, `CLAUDE.md`, `CHANGELOG.md`). Counters: tools 662 → 663 for the new module, which is allowlisted in `smoke_exceptions.txt` alongside its siblings.

Re-verified after the change: all four CLIs pass `--help`/`--sample`; the end-to-end pipeline runs clean; every blocking gate green (`compileall`, `check_paths --all`, `check_dual_publish`, `smoke_scripts` 0 failed, `derive_counters --check`, `check_plugin_json --all` 0 FAIL); security auditor unchanged at 0 critical / 4 high.

Thanks for the second pass — findings 1 and 2 were both places where this PR's own stated threat model had a hole in it, which is exactly the gap that's hard to see from the inside.

---

alirezarezvani:
All four addressed in `058447e`. Finding 1 was the best catch of the three reviews — it's a hole in a guarantee this PR makes in writing.

## 1. [High] Sniff path bypassed the zip budget — fixed

Confirmed exactly as described. `zip_safety.py`'s own docstring says *"every read goes through `safe_read()`... so a bomb is refused rather than materialized"* — and this one didn't, which makes it a documentation defect as much as a security one. Worse than the DOCX/EPUB cases too: sniffing runs **before** a format is even chosen, so it's the earliest attacker-controlled point in the pipeline.

Routed through `safe_read()`. One deliberate detail: its `ExtractionError` sits *outside* the surrounding `except (BadZipFile, KeyError, OSError)` tuple, so a bomb reports as a bomb rather than being swallowed into the generic "unsupported ZIP-based format" message.

```
$ # extensionless file, PK header, single `mimetype` member: 203,866 -> 209,715,220 bytes
$ extract_document.py sniffbomb
WARNING: skipping sniffbomb: archive member 'mimetype' declares 209,715,220 bytes
uncompressed; the per-member limit is 67,108,864 bytes
peak child RSS: 15 MB
```

## 2. [Medium] `--author`/`--author-url` ignored in the printed entry — fixed

Right, and the reasoning was the convincing part: that snippet exists *specifically* to prevent hand-editing mistakes, so printing a different name than the manifest beside it defeats its own purpose. `_marketplace_entry()` now takes `author` and is called with the resolved value.

```
$ skill_plugin_emitter.py --skill-dir skill --dest out --author "Jane Doe" --author-url https://jane.example
marketplace entry author: Jane Doe
manifest author:          Jane Doe | https://jane.example
```

## 3. [Low] TOCTOU between the guard and `copytree` — closed

Worth noting the window was already narrower than a plain check-then-act: `copytree` runs with `symlinks=True`, so a link planted in that window is copied *as a link*, not dereferenced — no target content is read. But "ships a package containing a symlink" is still not acceptable for something distributable.

Now closed properly by re-walking what actually landed: if any symlink is present after the copy, the package is **deleted** rather than shipped.

Verified with a monkeypatched guard that plants a link the instant the check passes:

```
REFUSED: symbolic link(s) appeared in the source tree while it was being copied —
the package has been removed rather than shipped:
  planted.md
package removed: True | secret leaked: False
```

## 4. [Nit] Unconditional `"license": "MIT"` — addressed

Kept the license string itself (the package *scaffolding* — manifest, agent, command, README — genuinely is MIT; it's the compiled notes under `skills/` that follow the source's terms). What was missing is that the distinction only existed in README prose. Added `source.license_scope` to the manifest so a tool reading it alone sees the caveat, plus a comment at the assignment:

```json
"license_scope": "The top-level `license` covers this package's scaffolding only. The compiled notes under skills/ are derived from the source document and carry that work's terms; see source.rights_basis."
```

---

Recorded as deviations 19 and 20 (count synced across `plugin.json`, `CLAUDE.md`, `CHANGELOG.md`). All gates green: `compileall`, `check_paths --all`, `check_dual_publish`, `smoke_scripts` (0 failed), `derive_counters --check`, `check_plugin_json --all` (0 FAIL); all four CLIs pass `--help`/`--sample`.

Three rounds of review have now found three separate places where the stated threat model had a gap the implementation didn't cover — workdir, then EPUB/copytree, now the sniff path. That's a useful signal in itself about vendored code: the parts that *look* already-hardened are where the unhardened path hides.

---

PR #947, opened 2026-08-09
feat(engineering): add memory-engineering skill — engineer the forgetting, not the remembering

alirezarezvani:
## Summary

New `engineering/memory-engineering/` plugin, filling a real gap: the repo had no skill for **designing, pricing, or auditing an agent memory system**. The nearest neighbours all bound something else — `llm-wiki` maintains one vault, `skillopt-sleep` runs a consolidation loop, `agent-harness` bounds a task loop. This bounds a **store**.

The thesis it operationalizes: *a storer optimizes what a system remembers; a memory engineer optimizes what it forgets.* The keystone script refuses to pass a memory design that has no forgetting policy.

### Four stdlib scripts, one per lens

| Script | Lens | What it does |
|---|---|---|
| `memory_cost_profiler.py` | Stanford — *what does it cost?* | Construction vs query spend split, **cost per correct answer**, amortization ratio, construction-co-located-with-queries warning |
| `memory_architecture_picker.py` | Stanford — *which cost to pay?* | Scores the four paradigm families, disqualifies on hard constraints, **names the cost the choice makes you pay**, and refuses to pick when the top two tie (exit 2 + the tie-breaking question) |
| `memory_density_auditor.py` | Microsoft — *what's worth keeping?* | Classifies records **FACT / SKILL / LOG / PROSE**, near-duplicates via word-shingle Jaccard + union-find clustering, staleness + volatile-wording flags, knowledge density per 1k tokens. Runs on a real `--dir` **or** `--jsonl` |
| `forgetting_policy_linter.py` | Anthropic + **the gate** | 8 checks; **F1** (explicit forgetting rule) and **F4** (contradictions surfaced, never auto-merged) block at exit 4 |

### Evidence discipline — two source paraphrases corrected, not propagated

The four-lens framing is synthesized from [@N01ennn's *"How to be a Memory Engineer"*](https://x.com/N01ennn/status/2083971749079581120), but **every quantitative claim is re-cited to the primary source**, with per-claim confidence levels (the `andreessen` precedent). Two of the article's numbers do not survive checking as stated:

1. **The 47× energy figure** is the *spread across ten evaluated systems* (BM25 at 4,145 J per correct answer vs A-Mem/MIRIX at 115–197 kJ — a "28–47× premium"), **not** "two systems with identical accuracy split by 47 times." Corrected in `memory_cost_canon.md` §2.
2. **The 97% first-pass-error reduction** is **Rakuten's named, vendor-published customer testimonial** (Yusuke Kaji, at 27% lower cost / 34% lower latency) — not a controlled study, and not a general property of building memory this way. Corrected in `memory_control_and_governance.md` §4.

Primary sources confirmed during the build: [arXiv:2606.06448](https://arxiv.org/abs/2606.06448) (Omri et al.), [arXiv:2604.09852](https://arxiv.org/abs/2604.09852) (MEMENTO), [Microsoft Research PlugMem](https://www.microsoft.com/en-us/research/publication/plugmem-a-task-agnostic-plugin-memory-module-for-llm-agents/), [Anthropic managed-agents memory](https://claude.com/blog/claude-managed-agents-memory).

### Three classifier defects found during the build

Each would have produced garbage output on any real repository — all three surfaced only because the auditor was run against a **real** directory rather than its `--sample`:

- **Headings inside fenced code blocks split records.** A `# comment` in a ```bash fence became a record boundary. Fixed by masking fences before heading detection, on a length-preserving copy so original-text offsets stay valid — 258 phantom records → 107 on a real directory.
- **Short fragments matched trivially at 1.00 Jaccard.** Two 1-word blocks share identical shingle sets. Fixed with a 20-word floor for duplicate candidacy — eliminated 41 false-positive "duplicates."
- **Signal-less prose was labeled LOG**, firing `LOG_HEAVY` at 74% on a prose documentation folder. Now its own **PROSE** class with its own verdict and fix.

### Three further defects found in review, all fixed

| Commit | Defect | Why it mattered |
|---|---|---|
| `cb49e3b` | `find_duplicates()` reported *redundant copies* (k−1) under the name `records_with_a_duplicate` (k) | Could sit under the 15% `DUPLICATE_SHARE` threshold and produce a false negative on the finding the tool exists to catch. Now reports both, with union-find clustering — endpoint counting gets k−1 wrong on 3+ clusters |
| `5c0e531` | 2 of 4 `TIE_BREAKERS` keys were unreachable (lookup sorts the pair; two keys were authored in the other order) | The two dead ones were the *plausible* near-ties, so the authored tie-breaking question silently degraded to the generic fallback. Keys now normalized at import, with a collision guard that **raises** — verified by injecting a reverse-order duplicate |
| `38cf9be` | README prose counter left at 746 | `derive_counters.py --check` **passes with a deliberately wrong value there** — the line is outside gate coverage. Flagged as a follow-up rather than widening shared infra inside a skill PR |

Every one of these was a *silent wrong answer*, not a crash — invisible to `--help` / `--sample` / exit-code checks.

## Checklist

- [x] Scripts run with `--help` without errors (all 4)
- [x] No vendor-locked dependencies without open-source fallback (stdlib-only: `argparse`, `json`, `os`, `re`, `sys`, `datetime`)

## Type of Change

- [x] New skill

## Testing

```
--help            4/4 OK
--sample          4/4 OK (exit 2 = findings surfaced, as designed)
--output json     4/4 parse as valid JSON
stdlib-only       argparse datetime json os re sys
error paths       missing file → 3 · malformed spec → 3 · failing policy → 4
```

- **`skill_review_checklist_runner.py` → 6/6 PASS** (binding for post-v2.6.0 skills; required trimming SKILL.md to 100 lines and unifying "tool"→"script" terminology)
- All six blocking CI gates pass locally and in CI: `check_plugin_json`, `check_paths`, `check_dual_publish`, `smoke_scripts`, `smoke_json_output`, `derive_counters --check`
- `skill_security_auditor.py` on the skill directory → **PASS (0 critical, 0 high, 0 info)**
- Density auditor verified end-to-end against a **real** directory (`standards/`), not only its `--sample`
- Union-find cluster arithmetic verified at k=2/3/4 → participants=k, redundant=k−1
- The combined `assets/memory_design_spec.example.json` verified to feed all three spec-taking scripts

## Counters

This branch merged `dev` in `83f53ee` after `book-to-skill` landed mid-review, which consumed the deltas originally claimed here. Numbers were **re-derived from the merged tree**, not hand-adjusted:

**skills 363→364 · tools 663→667 · refs 746→750 · agents 103→104 · commands 118→120 · plugins 89→90**

`derive_counters.py --check` is green.


alirezarezvani:
Thanks — the `find_duplicates()` bug is real and is fixed in `cb49e3b`. One correction to the diagnosis, though, which changed the shape of the fix.

## The bug, precisely

`len(seen)` counting only `j` isn't arbitrary: for a fully-connected cluster of *k* mutually-duplicate records it equals exactly **k−1**, which is the number of *redundant copies you could delete*. That's a meaningful metric.

The actual defect is that it was reported under the name `records_with_a_duplicate` and rendered as *"N records have a near-duplicate"* — and that wording describes **participants (k)**, not redundant copies (k−1). Two valid numbers, one reported under the other's name.

So rather than `seen.add(i)`, both are now computed and reported:

| Field | Meaning |
|---|---|
| `records_with_a_duplicate` | every record with ≥1 near-duplicate — drives `duplicate_share` and the `DUPLICATE_BLOATED` threshold, matching the message wording |
| `redundant_copies` | participants minus one survivor per connected cluster — what you could actually delete |

Clusters are resolved with **union-find**, not by counting pair endpoints. That matters for your 3+ cluster case: a 3-record cluster emits pairs `(i,j)`, `(i,k)`, `(j,k)`, so any endpoint-counting shortcut — including `seen.add(i); seen.add(j)` if you then tried to derive redundancy from it — gets the redundant count wrong.

Verified:

```
cluster k=2: pairs=1 participants=2 redundant=1
cluster k=3: pairs=3 participants=3 redundant=2
cluster k=4: pairs=6 participants=4 redundant=3
3-cluster + 1 unrelated: participants=3 redundant=2
```

Sample output moves from `1 record (17%)` to `2 records (33%), of which 1 redundant`. **Your threshold concern was correct** — this could tip a real store under the 15% `DUPLICATE_SHARE` line and produce a false negative on the finding the tool exists to catch.

## Other points

- **Stale type hint** — fixed. `-> list[dict]` removed; the 5-tuple return is documented in the docstring instead.
- **O(n²) scan** — capped at `MAX_DUPLICATE_SCAN = 2000` eligible records, and the tool now *prints how many were skipped* rather than truncating quietly. A silent cap reads as "no duplicates found," which this repo's no-silent-caps convention exists to prevent.

## Citations — verified, since you flagged you couldn't

Reasonable thing to escalate given the PR leans on evidence discipline. These were checked against primaries during the build, not asserted:

- **arXiv:2606.06448** → *Agent Memory: Characterization and System Implications of Stateful Long-Horizon Workloads* (Omri, Gan, Broveak, Geens, He, Pentland, Verhelst, Weissman, Tambe). Confirmed: four axes (construction/storage/retrieval/mutability), four paradigm families, ten evaluated systems, BM25 at 4,145 J/correct answer, *"the spread across agent memory systems exceeds 47×"*, *"none of the evaluated systems prune or forget by default"*, up to 9× footprint variance at 1M tokens.
- **arXiv:2604.09852** → *MEMENTO: Teaching LLMs to Manage Their Own Context*. Confirmed: ~2.5× peak KV-cache reduction, ~1.75× vLLM throughput, 15-percentage-point AIME24 drop when the KV shadow channel is removed, `microsoft/OpenMementos` (228K traces).
- **Rakuten 97% / 27% / 34%** → confirmed verbatim on Anthropic's *Built-in memory for Claude Managed Agents* post, attributed to Yusuke Kaji, GM AI for Business. Deliberately labeled in `memory_control_and_governance.md` §4 as a **named vendor-published customer testimonial, not a controlled study** — the source article presents it as a general property of the approach, and that framing is corrected rather than repeated.

The two corrections in the references (the 47× being a ten-system spread rather than an accuracy-matched pair; the 97% being one customer's result) are the two places the source article's paraphrase didn't survive checking.

All six blocking gates, `--help`/`--sample`/`--output json` on all four scripts, and the write-a-skill checklist (6/6) re-run green after the fix.

---

alirezarezvani:
Confirmed and fixed in `5c0e531` — this was a real bug and a good catch, because it fails silently.

## `TIE_BREAKERS` key ordering

Verified by execution before fixing:

```
key                                  sorted?  reachable
('flat_rag', 'structured_rag')       True     YES
('structured_rag', 'agentic')        False    NO  <-- DEAD
('long_context', 'flat_rag')         False    NO  <-- DEAD
('long_context', 'structured_rag')   True     YES

reachable: 2 of 4
```

Exactly as you described. And your point about *which* two die is the part that matters: `structured_rag`/`agentic` and `long_context`/`flat_rag` are the plausible near-ties, so the tool's most distinctive behaviour — an authored, specific tie-breaking question — degraded to the generic fallback with no error and no signal.

**Fix:** keys are normalized through `sorted()` at import, plus a collision check that raises if two entries describe the same family pair after normalization. I took your suggestion to add the self-check since the repo has no test suite — but made it *raise* rather than assert-and-continue, so a regression fails loudly at import instead of degrading quietly, which is the failure mode that hid this in the first place.

Verified by execution, not inspection:

- all 4 keys reachable after normalization;
- a constraint set that ties `structured_rag` against `agentic` now returns the authored question (*"Does your memory need to correct itself without a human in the loop?"*) instead of the fallback;
- injecting a duplicate pair in reverse order makes the guard fire — so the guard is tested, not merely present.

## SKILL.md step numbering

Fair — a reader skimming only the code block had no way to know step 5 existed. Added `# 5 - No command. Prove each pass by hand before scheduling it.` to the block, and compensated a line elsewhere to stay inside the write-a-skill checklist's 100-line limit (still 6/6 PASS).

## Note on the counter reconciliation

You mentioned the counters reconcile at 363→364 skills / 89→90 plugins. Worth flagging that those numbers moved *during* this PR: `dev` merged `book-to-skill` mid-review, which consumed the deltas this branch originally claimed. The branch merged `dev` in `83f53ee` and the headline numbers were re-derived from the merged tree rather than hand-adjusted — `derive_counters.py --check` is green, and the release note's own delta line was rewritten to sit on top of `book-to-skill` instead of asserting a stale range.

All six blocking gates, the four scripts' `--help`/`--sample`/`--output json` contract, and the checklist re-run green after both fixes.

---

alirezarezvani:
Confirmed and fixed in `38cf9be`. You were right — README.md line 30 was the only one left at 746.

**Why the sync missed it:** the pass replaced the string `"746 reference guides"`, which is the phrasing in `CLAUDE.md` and `marketplace.json`. Line 30 reads `"746 templates, checklists, and domain-specific knowledge files"` — different wording, never matched.

**The more useful finding is why CI didn't catch it.** I set that line to a deliberately wrong `111` and re-ran the gate:

```
derive_counters.py --check  ->  PASSES with 111
```

So this prose line is **outside the checker's coverage entirely**. That's the actual reason a reviewer found it and the automated gate didn't, and it means the same drift can recur silently on any future PR.

I did **not** widen `derive_counters.py` here. It's shared infra, and extending its coverage would likely surface pre-existing drift in unrelated rows and turn CI red for reasons unconnected to this PR — that's a maintainer call, not something to slip into a skill PR. Flagging it as a follow-up: `scripts/derive_counters.py` currently checks the `**Python tools** — N CLI scripts` line but not the `**Reference docs** — N templates...` line directly beneath it.

Also swept the rest of README for any stale counter adjacent to counter vocabulary (old value vs new, across skills/tools/references/agents/commands/plugins) — no other occurrences. Badges and both prose lines now read 364 / 667 / 750 / 104 / 120 / 90.

One line deliberately **not** changed: `CLAUDE.md` line 183 records `book-to-skill`'s own delta `refs 741 → 746`. That's historically accurate for that release note and shouldn't be re-pointed at the current total.

Thanks also for independently verifying the offset-preserving fence masking and the union-find cluster accounting — those were the two subtlest pieces, and both were fixed mid-review from earlier passes on this PR (the union-find rewrite landed in `cb49e3b` in response to a duplicate-undercount finding, and the tie-breaker normalization in `5c0e531`).

All six blocking gates pass.

---

alirezarezvani:
Both bugs reproduced and fixed in `2c384a9`. You flagged that you couldn't run anything — I ran both, and **the F1 finding is the most serious defect found on this PR.**

## 1. F1 bypass — confirmed, and worse than "worth tightening"

Reproduced exactly as you described:

```
{"rule": "asdf"}               -> F1=PASS  detail: "Forgetting is designed: ."
{"rule": "ttl"} (no ttl_days)  -> F1=PASS  detail: "Forgetting is designed: ."
```

A **misspelling silently passed the one check this entire skill exists to enforce.** The skill's own hard rule #4 is *"never call a design done without a forgetting rule"* — the gate violated its own premise, and the only hint was a nonsensical trailing `": ."`.

Fixed by inverting the logic to allowlist-based: **PASS is now unreachable unless a concrete mechanism is actually found** (`ttl_days > 0`, `max_records`/`max_bytes > 0`, or a decay setting), rather than inferring safety from what the rule *isn't*. Failure messages distinguish an unrecognized rule from a declared-but-unconfigured one, so a typo is never mistaken for a deliberate decision not to forget. Booleans are rejected where a number is expected; `ttl_days: 0` counts as absent.

Verified across 10 cases:

```
typo rule 'asdf'          F1=FAIL exit=4      -- valid: ttl 365           F1=PASS
'ttl' but no ttl_days     F1=FAIL exit=4      -- valid: capacity only     F1=PASS
'ttl' with ttl_days=0     F1=FAIL exit=4      -- valid: decay             F1=PASS
'capacity' no max         F1=FAIL exit=4      -- valid: ttl_days no rule  F1=PASS
empty forgetting          F1=FAIL exit=4
rule none                 F1=FAIL exit=4
```

Plus an assertion that the empty-mechanism string can no longer be emitted at all.

Your note that the existing fixtures don't exercise this path was the key insight — `SAMPLE_POLICY` has a fully-specified TTL and `FAILING_SAMPLE` has `"rule": "none"`, so `--sample`/`--sample-failing` bracket the two *well-formed* cases and never touch the malformed middle. That's a fixture-design lesson, not just a code bug.

## 2. `--print-sample-spec` unreachable — confirmed

All three scripts exited **2** with a usage error when the flag was passed alone. Your diagnosis of the mechanism is exactly right: argparse enforces a `required=True` mutually-exclusive group during `parse_args()`, before the `if args.print_sample_spec:` branch can run.

This broke the **first line of the workflow SKILL.md documents verbatim**:

```bash
python scripts/memory_cost_profiler.py --print-sample-spec > workload.json
```

Fixed with `required=False` plus explicit post-parse validation (your suggested shape). Verified the full round-trip on all three — generate the spec to a file, then feed it back in:

```
--print-sample-spec standalone   exit=0, valid JSON  (all 3)
round-trip generate -> consume   exit=2 findings     (all 3)
no-args                          helpful error naming every valid entry point
```

You also correctly identified *why* it slipped: the PR checklist covered `--help`, `--sample`, and `--output json` — but never ran `--print-sample-spec` standalone. A documented command that was never executed.

## 3. Nit

Identity dict in `render()` removed.

## On the citations

Already verified during the build, not asserted — and I'd posted the specifics [earlier in this PR](https://github.com/alirezarezvani/claude-skills/pull/947#issuecomment-5229835520): arXiv:2606.06448 (Omri et al. — the 4,145 J baseline, the 47× spread, *"none of the evaluated systems prune or forget by default"*), arXiv:2604.09852 (MEMENTO — ~2.5× KV reduction, the 15-point AIME24 drop), and the Rakuten 97%/27%/34% figures verbatim from Anthropic's post, attributed to Yusuke Kaji. The two corrections in the references are precisely where the source article's paraphrase didn't survive that check.

---

Re-verified after both fixes: 4/4 scripts `--help`/`--sample`/`--output json`; error paths 3/4/4; all six blocking gates; checklist 6/6 PASS; security auditor PASS (0 critical, 0 high, 0 info).

Worth recording that **all six defects found on this PR were silent wrong answers, not crashes** — a duplicate undercount, two dead tie-breakers, a stale counter outside gate coverage, a bypassable blocking gate, and an unreachable documented flag. None would have been caught by `--help`/`--sample`/exit-code checks, which is the whole verification surface this repo's conventions provide.

---

PR #948, opened 2026-08-09
feat(engineering): human-gate — audit of petergyang/human-review + batched human review as a verification artifact

alirezarezvani:
## Summary

Audits [`petergyang/human-review`](https://github.com/petergyang/human-review) and ships a conceptual derivation that fits this repo's stdlib-only conventions.

**First, a naming correction.** Despite how it reads, `human-review` is **not a humanizer** — it is a human-in-the-loop *visual review* harness. This repo already has two actual humanizers (`engineering/behuman`, `marketing-skill/content-humanizer`) and neither overlaps. The new plugin is about human **approval**, not human **voice**.

### The audit — `audit/human-review-2026-08/AUDIT.md`

Upstream is genuinely well-engineered. **Its own test suite passes 90/90** (run during the audit), and its security model beats most local-server tools: loopback-only bind, DNS-rebinding `Host` check, constant-time token compare, lexical **and** realpath traversal guard, a deliberately inert Markdown renderer, 45-minute idle self-shutdown.

**Verdict: do not vendor, do adopt the pattern.** It is ~5,200 LOC of Node 20 with an npm runtime dependency — failing the same stdlib-only test that kept the heavier `skillopt` package out in v2.11.2. The `loop-library` / `skillopt-sleep` vendoring precedent does not transfer, because those were vendorable *precisely because* they were stdlib-only.

Seven findings; the three material ones:

| | Sev | Finding |
|---|---|---|
| **F1** | 🔴 HIGH | The skill instructs the agent to run unpinned `npx -y human-review` — every invocation may fetch and execute a **newly published** version, unapproved. |
| **F2** | 🟠 MED | "Do not end your turn while it is waiting" + re-poll on timeout, with **no headless guard and no retry cap**. On CI or SSH this loops with no terminating condition — the exact AR5 loop-discipline gap `audit/engineering-agentic-2026-07/` already named as repo-wide. |
| **F3** | 🟠 MED | Only `/api/*` is token-gated; `/artifact/<key>` and `/s/<id>` are not. |

### The build — `engineering/human-gate`

The transferable idea is **structured human feedback as a machine-parseable verification artifact rather than chat prose** — the lane `agent-harness`'s `loop_controller.py` lacks, since it verifies by machine only and cannot represent "a person approved this."

Three stdlib-only scripts. No server, no socket, no network fetch:

- **`review_page_builder.py`** — Markdown/HTML → single-file anchored review page. **Zero network requests** (no CDN, no fonts, no Prism); ~11 KB, opens over `file://`. Escapes before applying inline markup, scheme-allowlists every href, drops `<script>`/`<style>`/`<head>` on HTML input.
- **`feedback_parser.py`** — sidecar → `batch.v1` JSON. BLOCKER/MAJOR/MINOR/NIT (matching `markdown-html/md-review`) plus EDIT/NOTE/APPROVE. Verifies every quote against the real file; strips HTML comments so a documented example cannot parse as a real sign-off.
- **`human_gate.py`** — `open`/`status`/`collect`/`close`/`reset`, atomic writes, `0700`/`0600` state. **G1–G6** refuse to close on: no collected round · an open BLOCKER/MAJOR · an unnamed reviewer · a sidecar changed after collection · an exhausted round cap (exit 5 = escalate, never pass) · an undocumented waiver.

**Loop discipline deliberately inverts upstream:** no blocking poll, a headless guard (`CI`/SSH/no `DISPLAY`), a round cap that escalates. The sidecar is hand-writable Markdown, so the loop still closes over SSH and in CI where no browser exists. The optional bridge to upstream's richer editor is opt-in, asked-first, and **always version-pinned** (`npx -y human-review@0.6.0`).

## Checklist

- [x] Scripts run with `--help` without errors (all 3, plus `--sample`)
- [x] No vendor-locked dependencies without open-source fallback — stdlib only, zero dependencies

## Type of Change

- [x] New skill
- [x] Documentation (public audit record)

## Testing

| Check | Result |
|---|---|
| Upstream `npm ci && npm test` (audit evidence) | **90/90 pass**, 0 fail |
| All 3 scripts `--help` / `--sample` | exit 0 |
| Import surface | stdlib only: `argparse hashlib html json os re subprocess sys time` (+ inline `importlib`, `webbrowser`, `tempfile`, `shutil`) |
| Generated review page | 0 external URLs, 0 `<script src=>`, 0 `<link href=>`, 11.4 KB |
| Gate rules G1–G6 | each refusal triggered and verified individually; G5 exits 5 on `open` and `collect` |
| End-to-end (`human_gate.py --sample`) | open → refused (G1) → collect → refused (G2) → collect → **GATE PASSED** |
| JS→parser round-trip | review page's `sidecar()` executed under node; output parsed back with 0 problems, EDIT `after` preserved verbatim |
| `scripts/check_plugin_json.py --all` | OK |
| `scripts/derive_counters.py --check` | passed |
| write-a-skill 6-item checklist | **6/6 PASS** |
| `skill_description_validator.py` | PASS |
| `skill_security_auditor.py` | 0 critical, 0 high beyond two guarded deletes + the known plugin-root `SKILL.md` path artifact (same finding `weekly-review` reports) |

**Two real bugs found and fixed during the build**, both by my own test material:
1. HTML input collapsed to a single block — `<html>` incremented depth before `<body>` opened, so no child ever saw depth 0. Fragments with no `<body>` also produced nothing.
2. HTML comments were not stripped from the sidecar, so a `## APPROVE` written as an *example inside a comment* in the shipped asset parsed as a **real sign-off** — precisely the silent false approval this plugin exists to prevent.

## Counters

skills 362 → 363 · tools 644 → 647 · refs 741 → 744 · agents 102 → 103 · commands 116 → 117 · plugins 88 → 89 · engineering row 84 → 85 (derived via `scripts/derive_counters.py --check`)

## Attribution

MIT © 2026 Peter Yang. **No upstream code is copied** — `plugin.json`'s `attribution.derivation_note` states per-finding what was taken (the pattern), what was built differently, and why.

---

https://claude.ai/code/session_01233Eggb2cjSYf96X6C3pCm

---
_Generated by [Claude Code](https://claude.ai/code/session_01233Eggb2cjSYf96X6C3pCm)_


alirezarezvani:
Both findings were real. Reproduced each before fixing; pushed in `daa4dde`.

## 1. Gate-coverage gap — fixed as **G7**

You were right, and it was worse than advisory-noise: it defeated the skill's whole premise. Reproduced first:

```
## BLOKCER b2
> We expect a 40% lift.
No source. Do not ship this.
```

```
Verdict: 0 blocking item(s) open
  [NIT] @b2 f1 — No source. Do not ship this.
  Problems (1): unknown severity/kind 'BLOKCER' ...

GATE PASSED — p.md
exit=0   ← a reviewer's genuine blocker, lost to one transposed letter
```

`cmd_close` now refuses while the last collected round carries unresolved integrity problems, covering all three classes you named:

| Case | Before | After |
|---|---|---|
| `## BLOKCER` typo | exit 0 | exit 2 — `G7 round 1 integrity: unknown severity/kind 'BLOKCER' ...` |
| `EDIT` with no `+ after:` | exit 0 | exit 2 — `G7 ... an edit without replacement text cannot be applied` |
| Quote not in target file | exit 0 | exit 2 — `G7 ... item f1 quotes text not found in p.md` |
| Clean sidecar | exit 0 | exit 0 (no false refusal) |
| Missing reviewer | G3 | G3 alone — **not** double-reported |

That last row is why I didn't just refuse on any non-empty `problems`: the parser already emits "no `reviewer:` line" and "APPROVE alongside blocking items", which G3 and G2 own. A small `GATED_ELSEWHERE` prefix filter keeps each failure reported once, by the rule that actually owns it.

I kept `collect` at exit 0 — collecting *succeeded*; the gate is where refusal belongs, and `--waive` already covers G7 like any other rule. G7 is documented in the script docstring, SKILL.md, README, the command, `plugin.json` (description + `derivation_note`) and CHANGELOG.

## 2. Stray artifact — removed, and made non-recurring

Confirmed tracked, and your trace of its origin was exactly right. Specifically: it came from the post-merge verification sweep, where I ran the `--sample` loop with cwd at the repo root, and `git add -A` then swept it into the merge commit.

Deleted. `.gitignore` now covers `*.review.html` and `.human-gate/`, which also protects anyone *using* the skill from committing a disposable review page. The sidecar (`<artifact>.review.md`) is deliberately **not** ignored — that is the reviewer's feedback and belongs in git next to the artifact.

## 3. Not acted on

`cmd_status`'s terse success line — you flagged it as "not a real issue" and I agree; the JSON branch already carries `blocking_open` for anything programmatic.

## Re-verified after the change

`derive_counters --check` passes · `check_plugin_json --all` 90/90 · write-a-skill checklist **6/6 PASS** · description validator PASS · all three scripts `--help`/`--sample` green · `human_gate.py --sample` still walks refuse → refuse → pass.

Thanks — G7 is a better gate than what I shipped, and it came from the review.


---

alirezarezvani:
All four were real. Reproduced each before fixing; pushed in `70c908a`.

## 1 · HIGH — reviewed HTML executed in the review page

Your read of the code was exact. Repro before:

```
onerror       present in #doc: True
javascript:   present in #doc: True
onclick       present in #doc: True
```

`sanitize_attrs()` now drops `on*` handlers plus `srcdoc`/`srcset`, and runs `href`/`src`/`action`/`formaction`/`poster`/`cite`/`background` through the same allowlist the Markdown path always had. `_safe_href` gained two things while I was in there: it strips control characters *before* reading the scheme (so `java&#9;script:` can't smuggle one past), and it permits `data:image/*;base64` only for image attributes. `DROP_TAGS` now also removes `iframe`/`object`/`embed`/`frame`/`base`/`applet` — you flagged execution, but an `<iframe src="https://evil.test">` would also have quietly broken the page's no-outbound-request property.

After:

```
onerror  removed: True   javascript: removed: True   onclick removed: True
<iframe  removed: True   evil.test   removed: True
legit https link kept: True    legit relative img kept: True
```

## 2 · MEDIUM — rendered-vs-raw quote comparison

Correct, and sharper than it looks: **G7 turned this from advisory noise into a false refusal.** I introduced that regression one commit earlier.

`verify_quotes()` now matches against the raw source **or** a rendered-text projection — inline markup stripped for Markdown (`**`, `*`, `_`, `` ` ``, `[t](u)`, headings, blockquote markers), tags stripped and entities unescaped for HTML. Verified both directions, because a fix that just stops reporting would be worse than the bug:

| Case | Before | After |
|---|---|---|
| Quote spanning `**40%**`, `` `code` ``, `[a link](x)` | G7 refused | passes, no problems |
| Fabricated quote not in the doc | caught | still caught |

## 3 · MEDIUM — stale waiver

Reproduced exactly as you described: waived close → `--reopen` → clean round 2 → `GATE PASSED` still printing `WAIVED: round 1: reviewer unavailable`. `state["waiver"]` is now cleared whenever a close passes with zero refusals. Round 2 output is now just reviewer + rounds.

You framed this as "narrow," and it is — but it's the tool lying in its own ledger, which is the one place this skill can't afford to.

## 4 · LOW — overloaded exit 4

Fixed by making the blocked case return **2**, the same code `close` uses for "the gate would refuse." That keeps 4 meaning exactly one thing:

```
no sidecar         -> 4
feedback waiting   -> 3
collected+blockers -> 2
```

Documented in the script docstring, SKILL.md, the command, and README: **0 clear · 2 blocked · 3 collect · 4 nothing yet.**

## One thing you didn't flag, that your finding exposed

Investigating #1 showed a claim of mine was over-broad. README and SKILL.md said "zero network requests" flatly. True of the page's own chrome; **not** true of a reviewed HTML artifact's own `https:` images, which do load — and must, or the review isn't faithful. Both now say precisely that.

## Re-verified

`derive_counters --check` · `check_plugin_json --all` · write-a-skill checklist **6/6 PASS** · description validator PASS · all three scripts `--help`/`--sample` green · `human_gate.py --sample` still walks refuse → refuse → pass.

Two review rounds have now each found something real — G7, then the HTML sanitization gap. Both were in code I'd already called verified, which is the useful reminder.


---

alirezarezvani:
All three were real, and the first one was worse than "a bug" — the documented use case was **entirely non-functional**. Pushed in `4e59391`.

## 1 · HIGH — the HTML path was dead for real HTML5 documents

Your diagnosis was exactly right, including the reason it survived. Repro on your fixture, before:

```
$ review_page_builder.py landing.html
No reviewable blocks found in landing.html
exit=2
```

`meta`/`link`/`base` are void, so `html.parser` fires `handle_starttag` and never a matching `handle_endtag` — but they were in `DROP_TAGS`, so each bare `<meta charset>` incremented `_skip` permanently and every callback inside `<body>` hit the guard. Fixed as you suggested:

```python
if tag in DROP_TAGS:
    if tag not in self.VOID:   # void tags never fire handle_endtag
        self._skip += 1
    return
```

After: `3 blocks`, head correctly excluded.

**On why it survived three rounds** — you were right that the fixtures were the problem. Every HTML sample I wrote, including the one for the sanitizer pass, used `<title>`/`<style>` in head and never a bare `<meta>`. So I'd been testing a shape that doesn't occur in the wild while claiming the landing-page path worked.

Your regression suggestion is now the fix's own guard. `--sample` builds **both** fixtures, asserts expected block counts, and **exits 2 on regression**:

```
$ review_page_builder.py --sample
  quarterly-plan.md    6 blocks,  11.4 KB  ok
  landing.html         3 blocks,  11.1 KB  ok
```

It also writes to a temp dir now instead of cwd — the same mistake class that leaked a stray artifact in an earlier round.

## 2 · MEDIUM — `xlink:href`

Confirmed surviving the hardening added one commit earlier:

```
xlink:href="javascript:alert(1)"      → before
xlink:href javascript: STRIPPED       → after
```

Added `xlink:href`, `xlink:role`, `xlink:arcrole` to `URL_ATTRS`. Not deferring it to the audit doc as an accepted gap — you're right that closing it is cheaper than documenting why it's open.

## 3 · MEDIUM — `status` wasn't a real preview of `close`

Correct, and it contradicted a contract I'd written one commit earlier. Rather than duplicate the rule checks, I extracted `gate_refusals(state, sidecar)` and had both commands call it, so they can't drift when a rule changes:

| Scenario | `status` | `close` |
|---|---|---|
| Round with no reviewer, no blockers | **2** (was 0) | 2 |
| Clean approved round | 0 | 0 |

`status` now also prints *which* rules would refuse, not just a count.

## Standing back

Three review rounds, three real findings, each in code I had already called verified — G7, then the XSS gap, now a use case that never worked. The pattern is consistent: my fixtures kept testing the shape I had in mind instead of the shape that occurs in practice. The `--sample` regression check is the first thing in this PR that pushes back on that, which is why I wired it to fail loudly rather than just adding one more manual test.

## Re-verified

`derive_counters --check` · `check_plugin_json --all` · checklist **6/6 PASS** · description validator PASS · all three scripts `--help`/`--sample` green.


---

alirezarezvani:
All four were real. Pushed in `aac29fc`. Finding 3 was the most consequential thing found in any of the four rounds.

## 3 · The gate was one flag away from opt-out (taking this first — it's the big one)

You filed this as Low/Medium docs. It isn't. Reproduced:

```
$ human_gate.py close p.md --waive "no time"
GATE PASSED — p.md
  WAIVED: no time
          (over: G1 no review round has been collected — nobody has looked at this yet)
exit=0
```

A skill whose entire premise is "refuses to let an agent mark its own homework" could be fully bypassed with one flag and any string. And as you noted, that's the *most* tempting shortcut precisely when a human isn't available — the exact situation the tool exists for.

I didn't take the documentation fix. **G1 is now structurally unwaivable:**

```
$ human_gate.py close p.md --waive "no time"
GATE REFUSED — p.md
  ✗ G1 no review round has been collected (NOT WAIVABLE)

--waive accepts objections a reviewer raised; it cannot stand in for
review itself. Open a round and get one collected first.
exit=2
```

The principle: a waiver **accepts objections a reviewer raised**. It cannot manufacture a review that never happened. Waiving a genuine G2/G3/G4/G7 objection still works, verified.

## 1 · Inline `style` beacon

Confirmed — `background-image:url(https://attacker.example/beacon.png)` fired on open, no script involved, directly contradicting the property the README and manifest advertise. Added to `DROP_ATTRS`; now stripped.

Your framing sharpened it: the `<style>` **tag** was already dropped, so keeping the **attribute** was inconsistent as well as leaky. The page was already unstyled — inline style was buying nothing and costing the invariant.

## 2 · Markdown images

Confirmed, including the dead-code observation:

```
before:  <p>!<a href="foo.png" rel="noreferrer noopener">screenshot</a></p>
after:   <p>``&lt;img src="foo.png" alt="screenshot"&gt;``</p>
```

Added an `IMAGE` regex ahead of `LINK` plus a negative lookbehind on `LINK`, routed through the same allowlist — so `_safe_href(image=True)` is now actually reachable from the Markdown path, as its docstring always claimed. Verified `[x](javascript:alert(1))` degrades to inert alt text, not an attribute.

You also spotted that `feedback_parser.plain_text()` already had the `[alt](url) → alt` rule while the renderer didn't. That drift is what made this invisible: quote-matching behaved as if images rendered.

## 4 · Unterminated `<script>`

Confirmed — content after it silently disappeared. Per HTML parsing the swallow is arguably correct, so I kept the semantics and fixed the *silence*:

```
Warning: unterminated <script> in the source — everything after it was dropped.
Close the tag if that content should be reviewable.
```

## Nit · `rel` on raw-HTML `target="_blank"`

Taken: `<a href="https://x.test" target="_blank" rel="noreferrer noopener">`.

## Four rounds in

Each round found something real in code I'd called verified. The through-line is that my fixtures kept encoding my mental model instead of reality — head tags without `<meta>`, sentences without inline markup, waivers used only the way I intended them. The `--sample` regression gate from round 3 is the one thing so far that pushes back on that automatically; it can't catch a design hole like G1.

## Re-verified

`derive_counters --check` · `check_plugin_json --all` · checklist **6/6 PASS** · description validator PASS · all three scripts `--help`/`--sample` green.


---

alirezarezvani:
Both real, both fixed in `d6cff73`. One was worse than reported.

## 1 · `state_dir()` anchored to cwd

Reproduced exactly as described:

```
$ cd /repo && human_gate.py collect p.md     # round 1 recorded
$ cd /repo/sub && human_gate.py close ../p.md
GATE REFUSED — p.md
  ✗ G1 no review round has been collected — nobody has looked at this yet
```

You're right that it fails closed, but the cost isn't only confusion: it would push an agent into re-opening review rounds that already happened, burning a human's time twice and eating the round cap.

State now follows the artifact — the same anchor the sidecar and review page already use, so all three live together. `--state-dir` still wins. Verified: collect from the artifact's directory, `close` from a subdir → **GATE PASSED, rounds: 1**.

## 2 · Placeholder collision — worse than reported

You called it "corrupting the visible review page." It's more than cosmetic. Reproduced on a doc that mentions the tokens:

```
<p data-hg="b2">The template uses tok.md and {"target": "tok.md", "sidecar":
"tok.review.md", "round": 1, "blocks": [{"id": "b1", "text": "Doc"}, {"id...
```

The **entire JSON config** landed in the visible body.

Fixed with your third suggestion rather than reordering, because reordering only moves the problem: substituting `__CONTENT__` last would still break when a *block's text* — which is embedded in the config JSON — contains `__CONTENT__`. All three slots now fill in one `re.sub` pass, so no substituted value can ever be re-substituted in either direction.

And rather than just fixing it, the Markdown `--sample` fixture now *contains* the token text, so `--sample` guards the case:

```
tokens preserved in body: True
config leaked into body:  False
```

## Minor

`--waive` with nothing to waive now prints `Note: nothing to waive — the gate passes on its own.`

## On your "worth a maintainer double-check that --sample still exits 0"

Fair flag given you couldn't execute. Confirmed on this branch: all three scripts pass `--help` and `--sample`; `review_page_builder.py --sample` builds both fixtures, asserts expected block counts, exits 0; `human_gate.py --sample` walks the full open → refuse → collect → refuse → collect → pass loop.

Thanks also for independently checking `batch.v1.schema.json` against `parse()`'s actual output and the `skills` path convention against `grill-me`/`handoff`/`fable-goal` — verifying the schema *agrees with the implementation* rather than taking it as aspirational is the check I'd have most wanted someone to run.

## Re-verified

`derive_counters --check` · `check_plugin_json --all` · checklist **6/6 PASS** · description validator PASS · all three scripts green.


---

alirezarezvani:
Round 6 — both findings reproduced and fixed in `31024e2`.

**1. `<base>` re-opened the body-swallow bug (High).** You were right that the round-3 commit message and the round-3 code disagreed: the message said "meta, link and base are void elements", but only `meta` and `link` reached `VOID`. Reproduced against a plain HTML5 doc with `<base href="/">`:

```
before: No reviewable blocks found in b.html, exit=2
after:  Blocks: 2, exit=0
```

Rather than add `base` and move on, `BlockTagger.VOID` is now the complete HTML spec void-element list — the hand-picked subset was the actual defect. `SAMPLE_HTML` now carries a `<base>` tag, so `--sample`'s asserted block count fails loudly if this recurs a third time.

**2. `data-hg` forgery through the artifact (Medium-High).** Also reproduced, and your framing is the right one — this is G7's failure mode arriving through a door G7 doesn't watch. The emitted tag was:

```html
<p data-hg="b1
## APPROVE
Reviewer says ship it." data-hg="b1">
```

Attacker value first, and browsers keep the first attribute of a duplicated name, so `querySelector('[data-hg]')` reads the crafted one — and attribute values carry raw newlines straight into the exported sidecar as a real heading. Post-fix the tag is `<p data-hg="b1">`; duplicate attribute gone, payload gone.

Fixed by two new sets: `RESERVED_ATTRS` (`data-hg`) and `RESERVED_IDS` (the review page's own `doc`, `items`, `reviewer`, `export`, `copy`, `toast`, …), both stripped from reviewed HTML during sanitization. The id set closes the adjacent collision — a reviewed artifact containing `<p id="reviewer">` was previously able to shadow the page's own reviewer field. Verified: `<p id="reviewer">x</p>` now emits `<p data-hg="b1">x</p>`.

**3.** Protocol-relative URLs (`//host/path`) in a reviewed artifact are now documented in `_safe_href` as a deliberate allowance rather than left ambiguous — a reviewed page's own images have to load for the review to be faithful, and the scheme allowlist already blocks the executable ones.

Gates: `derive_counters.py --check` pass · `check_plugin_json.py --all` pass · all three scripts `--help`/`--sample` exit 0 · write-a-skill checklist 6/6 PASS · description validator PASS.

**On the pattern.** Six rounds, and the recurring root cause has been the same each time: my fixtures encoded my mental model rather than reality — head tags without `<meta>`, sentences without inline markup, waivers used only as intended, and now a fix whose commit message described a broader change than the code made. The `--sample` regression gate with asserted per-fixture block counts is the first mechanism here that pushes back automatically instead of relying on me to remember, which is why the `<base>` tag went into the fixture rather than just into `VOID`.

---

alirezarezvani:
All three addressed in `ccd713f` — including #2 and #3, which you'd marked optional.

**#1 — `marketplace.json` overclaim.** Confirmed and fixed. `aece872` corrected this wording in README.md, SKILL.md and plugin.json and missed marketplace.json; it now reads the same qualified form.

**#2 — cross-file prose coupling.** Worth doing, because the failure mode is reproducible rather than theoretical. The prefixes now live in `feedback_parser.py` beside the `problems.append()` calls that emit them (`P_NO_REVIEWER`, `P_APPROVE_WITH_BLOCKING`, exported as `GATED_ELSEWHERE`), and `human_gate.py` reads them off the loaded parser module — it already imports it, so this costs nothing. The old literal stays as a fallback for a parser too old to export the names.

I reworded the G3 message on pre-fix code to confirm your prediction:

```
✗ G3 round 1 has no named reviewer
✗ G7 round 1 integrity: unknown severity/kind 'BLOKCER'
✗ G7 round 1 integrity: the sidecar names no reviewer — ...     ← duplicate
```

Same reword against the fixed code produces only the first two lines.

**#3 — `ATTR_RE`.** I fixed the parse rather than only noting it, since it was one line: `target="q3 plan.md"` used to yield `('target', '"q3')`, and now yields `q3 plan.md`. Single/double quotes both work. The docstring states the quoting rule and, as you noted, that `target` is a display hint — verification runs against the `--target` path.

On the two you confirmed-and-cleared: agreed on both. G4's silent no-op when the sidecar is deleted post-collect is intentional — the collected round is the evidence, and the sidecar is a transport. If it needed a rule it would be its own, not G4's.

Gates: `derive_counters.py --check` pass · `check_plugin_json.py --all` 90/90 OK · marketplace.json parses · all three scripts `--help`/`--sample` exit 0.

---

alirezarezvani:
Thanks — two of the three led to changes in `5d03ed9`; the third I'm deliberately leaving as-is with reasoning below.

**`--max-rounds` not persisted (your "minor design note").** This was the most substantive of the three and it's now fixed. You were right that the cap was convention-dependent: `max_rounds` was only ever written into the `status` *payload*, never into gate state, so every command re-derived it from its own flags.

```
before: open --max-rounds 2  →  status  →  "max_rounds": 5
after:  same sequence         →  "max_rounds": 2
```

`open` now pins the cap in state and status/collect/close inherit it. Passing the flag again is treated as an explicit renegotiation and announces itself (`Round cap changed for this artifact: 2 -> 9.`) rather than silently overriding — a cap you can change by accident is the same problem in a different shape.

**Protocol-relative URLs.** Took the fixture route over the doc note, since your framing is the right one: the void-element and `__CONFIG__` regressions each earned a fixture, and this allowance is as much a decision as those were. `--sample` fixtures now carry must-keep / must-drop URL assertions next to the block count, and `SAMPLE_HTML` exercises all three cases — `//cdn.example.com/pricing`, an `https:` image, and `javascript:alert(1)`.

Verified the assertion actually bites rather than just passing by construction — adding `"javascript"` to the scheme allowlist:

```
landing.html   4 blocks, 11.4 KB  FAIL — kept javascript:
Sample regression: landing.html          exit=2
```

README updated so the prose and the fixture say the same thing.

**`<svg><use href="https://...">` — no change, and I don't think there should be one.** `href` is already in `URL_ATTRS` and scheme-checked; `https:` passes there under exactly the same deliberate rule that lets a reviewed page's own ``'&lt;img src="https://..."&gt;'`` load. Blocking cross-origin `use` while allowing cross-origin `img` would be inconsistent without buying anything: both are a network fetch, neither executes, and the fetch is the accepted tradeoff for a faithful review. If the no-fetch guarantee ever needs to be absolute, the change is dropping `https:` from `URL_ATTRS` entirely — a different and bigger decision than special-casing one element.

Gates: `derive_counters.py --check` pass · `check_plugin_json.py --all` pass · all three scripts `--help`/`--sample` exit 0 · write-a-skill checklist PASS.

---

PR #994, opened 2026-08-25
feat(engineering): deep-learning-book — companion skill for the free Goodfellow/Bengio/Courville textbook

alirezarezvani:
## Summary

Requested as "convert https://www.deeplearningbook.org/ into a skill". Shipped as a **companion**, not a compilation — and the reasoning is the reusable part of this PR.

**Why `book-to-skill` was not used.** Its emitter refuses a `shareable` package without `public-domain` / `open-license` / `internal-docs` / `author-permission`. None applies to an MIT Press title whose own site states the HTML-only format exists as a friction against copying under the authors' contract, and whose PDF distribution is not permitted. Its `rights_and_provenance.md` lists "publish a compiled skill of a copyrighted book to a public marketplace" under **Do not**, and its hard rule 1 forbids scraping a book from the web, so the pipeline could not have run against a URL in any case.

**The rule this sets for the repo** (recorded in `references/rights_and_use.md`): when a user asks to convert a copyrighted work into a *shareable* skill, build a companion that indexes and updates the source, not a compilation that reproduces it. Compile only when the gate clears; keep the output local when it does not.

**What shipped.** The compiled-skill *shape* — master `SKILL.md` (~2.0k tokens, chapter index + topic index), `chapters/ch01..ch20`, `glossary.md`, `patterns.md`, `cheatsheet.md` — filled with original synthesis. No passages, figures, exercises, or per-paragraph paraphrase; every chapter file links to the official free chapter. The organizing structure is the book's published table of contents.

**The differentiator is the delta layer.** A compilation freezes a source at its publication date; this one dates it. Every chapter carries a "What changed after 2016" section, and `references/book_to_2026_delta.md` gives five corrections with primary citations and per-claim confidence levels:

1. **Double descent** qualifies Ch 5's U-curve — "shrink the model when it overfits" is no longer the only correct move, and `capacity_planner.py` encodes the correction.
2. **AdamW** — weight decay ≠ L2 under an adaptive optimizer; Ch 7 treats them as interchangeable.
3. **Transformers** displaced Ch 10's recurrence — keep its gradient-flow analysis, drop its architecture recommendation.
4. **Diffusion** grew directly out of Ch 14's denoising autoencoders and Ch 18's score matching.
5. **Self-supervised learning** vindicated Ch 15's bet while replacing every method it names.

Two commonly repeated claims are marked contested rather than propagated (batch norm's "internal covariate shift" mechanism; the strong lottery-ticket form) and two are named as folklore. Stated rule: **the conflict is almost always in the recommendation, not the analysis.**

**Four stdlib tools, each with a real refusal:**

| Tool | Refuses |
|---|---|
| `reading_path_planner.py` | Exit 3 on a goal the book does not cover (names what does); exit 4 with forcing questions when unroutable |
| `training_diagnostics.py` | Exit 4 rather than diagnosing with no instruments; rules fire in priority order so a NaN is never reported as overfitting |
| `capacity_planner.py` | Exit 4 on a val-below-train split (leaky measurement); ranks "shrink the model" **last** when overparameterized |
| `model_arithmetic.py` | Exit 5 naming the layer whose shapes do not connect; exit 4 on an unknown layer type |

Plus 4 references citing 7–8 sources each, 3 assets, `cs-deep-learning-tutor`, and `/cs:deep-learning` · `/cs:dl-reading-path` · `/cs:dl-diagnose`.

**Note on the diff:** this branch was cut from `main`, which currently sits 3 commits ahead of `dev` (the `linkedin` plugin, merged via #991). Those commits ride along in this PR against `dev`; the work under review here is the single `deep-learning-book` commit.

## Checklist

- [x] **Target branch is `dev`** (not `main`)
- [x] Scripts run with `--help` without errors — all four also support `--sample` and `--output json`
- [x] No hardcoded API keys, tokens, or secrets — no network calls anywhere in the plugin
- [x] No vendor-locked dependencies without open-source fallback — standard library only
- [x] Follows existing directory structure (`engineering/deep-learning-book/skills/deep-learning-book/SKILL.md`)

## Type of Change

- [x] New skill
- [ ] Improvement to existing skill
- [ ] Bug fix
- [x] Documentation (counters + release notes in `README.md`, `CLAUDE.md`, `marketplace.json`)
- [ ] Infrastructure / CI

## Testing

Every gate run against the repo's own tooling:

- `book_skill_validator.py` (from `book-to-skill`, all four check families) — **passed all checks, exit 0**
- `token_budget_estimator.py --skill-dir` — every file inside its cap; resident core 2,010 / 4,000 tokens, 20 chapters averaging ~644 tokens each
- `check_plugin_json.py --all` — `OK engineering/deep-learning-book/.claude-plugin/plugin.json`; marketplace description 899 chars (cap 1,024)
- `derive_counters.py --check` — **passed**: skills 386→387, tools 723→727, refs 838→842, agents 116→117, commands 146→149, plugins 97→98
- All four tools: `--help`, `--sample`, and `--sample --output json` (valid JSON) verified; typed exit codes confirmed on the refusal paths (3, 4, 5) and the success paths
- `model_arithmetic.py` arithmetic hand-checked against the sample convnet (conv 3×3×3×32+32 = 896; linear 4096×128+128 = 524,416; MHA 4·768²+4·768 = 2,362,368; LSTM 4·(768·256+256²+2·256) = 1,050,624)
- `model_arithmetic.py --spec assets/example_layer_spec.json` — runs clean

---
_Generated by [Claude Code](https://claude.ai/code/session_01BswsZp5zrJWFAGU6KWNA1s)_


alirezarezvani:
Both code nits verified against the source and fixed in `c75500f`.

**1. `include_optional` naming** — confirmed: the `plan()` parameter was `include_optional` while the flag and call site both used `include_intro`, and it only ever gates ch01. Renamed the parameter and its use. Behavior unchanged, checked both directions: `--include-intro` keeps ch01 first, the default drops it.

**2. conv2d `"same"` padding convention** — confirmed, and a fair catch: `ceil(H / stride)` is the TensorFlow/Keras SAME convention and the tool disclosed nothing. Now documented in the module docstring, including that PyTorch's `padding='same'` is symmetric-only and rejects a stride other than 1, so a strided PyTorch layer will not match this row — with `"valid"` plus explicit padding named as the exact-case workaround. Added a pointer comment at the computation itself. No arithmetic change; the sample stack reports identical parameters and FLOPs.

**3. Gates you couldn't execute** — no code change needed, but for the record, all of these are green locally on the current head: `compileall`, `check_paths`, `check_frontmatter`, `check_dual_publish`, `check_model_freshness`, `smoke_scripts` (696 passed / 0 failed), `derive_counters --check`, `check_skill_names`, `check_plugin_json`, `book_skill_validator`, and `--sample --output json` on all four tools.

One unrelated thing surfaced while running the suite, left alone deliberately: the advisory `smoke_json_output.py` gate (G9, `continue-on-error`) reports 8 failures, all in `agent-launcher/` scripts that print non-JSON to stdout under `--sample --output json`. Pre-existing and outside this diff — worth a separate issue rather than widening this PR. All four tools here pass it.

Note that the first CI run on this branch was red on gate G1 for a real defect of mine (bare `references/`/`assets/` links in the agent and command files, which sit outside the skill directory and so resolved against none of the linter's three bases); fixed in `cf572c8`.

---

alirezarezvani:
Verified the lane tie-break observation and fixed the cause in `1366714`.

Reproduced first: `"train a transformer"` hits `practitioner` on `"train"` and `sequence` on `"transformer"` — one keyword each — and the alphabetical tie-break picked `practitioner`, exactly as described.

Rather than swapping the example in `SKILL.md`, `score_lanes()` now breaks equal-hit ties on **keyword specificity**: the lane whose longest matched keyword is longest wins, since a tie between a generic term and a discriminating one shouldn't be settled by luck. Lane key stays the final tie-break so ordering remains deterministic.

Regression battery over eight goals — `"train a transformer"` now routes to `sequence`, and the vision, generative, foundations, practitioner, complete, representation and sequence goals all route exactly as before. Refusal paths unchanged (out-of-scope exit 3, unroutable exit 4, sample exit 0).

Gates green after the change: `compileall`, `check_paths`, `check_frontmatter`, `check_dual_publish`, `check_model_freshness`, `smoke_scripts` (696 passed / 0 failed), `derive_counters --check`, `check_skill_names`, `check_plugin_json`, `book_skill_validator`, and `--help` + `--sample --output json` on all four tools.

One related case I looked at and deliberately left alone: `"why does my training diverge"` routes to `practitioner` rather than `optimization`, but that's a hit-count win (`"train"` and `"training"` both match), not a tie-break — and the `practitioner` lane already includes ch08. Changing it would mean reworking the keyword sets, which is beyond what this finding needs.

---

alirezarezvani:
Good catch — this was a real bug, and the diagnosis was exactly right. Fixed in `eeb3cb9`.

Reproduced first: the shipped asset reported **1,207,962,624** parameters for the FFN up-projection instead of **2,362,368**, off by exactly 512× the sequence length, putting the block total at ~1.21B instead of ~7.09M.

**Fixed the cause, not just the spec.** The root problem was in the tool: `mha` emits `(seq, d_model)` but `linear` refused *any* 2-D input, so the only way to attach a feedforward block was to flatten — which models a dense layer over the whole flattened sequence, a different layer with `seq_len` times the parameters. A transformer FFN was therefore not expressible at all, and the example walked straight into the trap the docstring warned about. `linear` on a 2-D `(seq, features)` input is now **position-wise**: one weight matrix shared across positions, parameters independent of sequence length, compute linear in it. The 3-D path still refuses (exit 5) with an updated message pointing at `flatten`. The asset drops the flatten and its comment now records why it must not come back.

Verified numerically rather than by exit code:

```
layernorm x2   3,072
mha            2,362,368   = 4*768^2 + 4*768
ffn up         2,362,368   = 768*3072 + 3072
ffn down       2,360,064   = 3072*768 + 768
total          7,087,872   <- matches the tool exactly, and a BERT-base encoder layer
```

Convnet `--sample` is unchanged at 545,098; `linear` on 3-D input still exits 5.

**CHANGELOG entry added** — correct, it was missing while `CLAUDE.md`, `README.md` and `marketplace.json` all carried the new skill and its counter deltas.

Your closing point is the one I'd most like to keep: exit-code correctness is not numeric correctness, and this slipped through precisely because `--spec` ran clean. I hand-checked the convnet sample when building the tool but not the transformer asset — which was the one with the subtle semantics. The per-layer arithmetic above is now in the commit message so the expected numbers are reviewable, not just the exit code.

---

alirezarezvani:
Thorough review — thank you. Findings **3** and **4** are against this PR's own code and are both fixed in `e923237`. The rest are against `marketing/linkedin`, which this branch carries but does not touch; details on that below.

## Fixed (findings 3 and 4)

**3 — substring matching in `reading_path_planner.py`.** Reproduced exactly: `--goal "train models for image storage and retrieval"` exited 3, confidently refused as out-of-scope RAG work (`"rag"` inside **sto·rag·e**), and `"an exploratory look at optimization"` exited 3 citing LoRA (`"lora"` inside **exp·lora·tory**). `"conv"` matched *converge* and `"text"` matched *context*. This was the worst class of bug for this particular tool — its stated design is to refuse rather than guess, and it was guessing with certainty.

Matching is now word-boundary anchored with an optional plural, plus an explicit surface-form table for the few tokens whose inflections a bare `\b` match would miss (`fine-tuning`, `prompting`, `agentic`). Verified after: both goals above route correctly, `"converge"` reaches the `optimization` lane rather than `vision`, and every real refusal still refuses — RLHF, LoRA fine-tuning, RAG pipelines, prompt/agent goals all still exit 3.

**4 — malformed input crashing past exit 4 in `model_arithmetic.py`.** Reproduced all five cases you named; each produced a traceback with exit 1. Numeric fields now go through a checked accessor rejecting non-integer and non-positive values, the input layer's `shape` is validated, and the top-level spec plus every layer entry are type-checked:

```
non-dict top-level JSON  exit 4  spec must be a JSON object, got list
non-dict layer entry     exit 4  layer 0 is str, not an object
"stride": 0              exit 4  layer 1 (conv2d): 'stride' must be positive, got 0
"groups": 0              exit 4  layer 1 (conv2d): 'groups' must be positive, got 0
"filters": "abc"         exit 4  layer 1 (conv2d): 'filters' must be an integer, got 'abc'
```

Regression battery over eight goals routes exactly as before; convnet sample still 545,098 parameters, transformer asset still 7,087,872. Full gate suite green.

## Not fixed here — and why

Findings **1**, **2**, **5**, **6** and most of the smaller items are in `marketing/linkedin/`. That plugin was reviewed and merged separately in #991; it appears in this diff only because this branch was cut from `main`, which sits ahead of `dev`. Fixing them here would put substantive changes to another plugin inside a PR about `deep-learning-book`, where they'd get no focused review and would be invisible in the PR title.

They look like real bugs from your traces, and two are worth prioritizing — the `P2-SCRAPING` rule refusing the very self-export its own substitute recommends, and `outreach_message_builder.py`'s ask-detection being bypassable via `--reason`, since that one contradicts a headline claim in the plugin's own description. I'd suggest a separate issue against `marketing/linkedin`; I'm happy to take it as its own PR if that's wanted.

Also worth recording: your note that `gh pr view --files` truncates at 100 entries while this PR touches 142 files is a good catch about the review surface itself, independent of any finding.

---

alirezarezvani:
Correct on both counts, and it's the one field I missed when hardening the others in `e923237`. Fixed in `8f833d2`.

Reproduced before fixing, and the second case is slightly worse than described:

```
seq_len "ten"   ->  uncaught ValueError, traceback, exit 1   (documented: exit 4)
seq_len 0       ->  no error at all: reported a plausible "0x8" layer with zero
                    activation elements and exited 0
```

So the non-positive case wasn't merely unvalidated — it presented a degenerate model as a valid one, which for an arithmetic tool is the worse of the two failures.

Routed through `_positive_int` with the existing default, preserving behavior: both now exit 4 naming layer and field, the default fallback still resolves to the incoming sequence length (`16x8`), and an explicit valid `seq_len` is still honoured (`32x8`).

Since the finding was really "one field escaped a convention the rest of the file enforces," I swept the file rather than patching the single instance: **all 13 numeric spec fields now go through `_positive_int`, and no raw `int()` on a spec field remains.** Reference specs unchanged — convnet 545,098 parameters, transformer asset 7,087,872. Full gate suite green.

Worth recording for anyone reading this thread later: this is the second bug found in the exact seam where a fix was applied — the 512× parameter error was in the tool's flagship example, and this one was in the field the hardening pass skipped. Both were in `model_arithmetic.py`, and both slipped past because the spec exited 0. Clean exit still isn't correct numbers.

---

alirezarezvani:
Checked the `--include-intro` note rather than just documenting it, and it was worse than reported: the flag was **inert on every lane, `complete` included** — not merely limited to `complete`. Fixed in `98019ac`.

Two causes compounded. ch01 is not a prerequisite of any chapter, so it never arrived through closure; and `complete`, the one lane that targets it, skipped the ch01 filter entirely. So the filter removed a chapter that was never present, and the flag controlling it could not change any output:

```
read the whole book              --include-intro -> NO CHANGE
convolutional models             --include-intro -> NO CHANGE
generative models and diffusion  --include-intro -> NO CHANGE
learn the math foundations       --include-intro -> NO CHANGE
```

Inverted the logic so the flag *adds* ch01 rather than un-removing it. It now has a real effect on every lane, `complete` stays unchanged (ch01 is already among its targets), and the help text says so. Verified per lane: `complete` unchanged; vision, generative and foundations each gain ch01 with the flag and are untouched without it; ch01 sorts first and never precedes a chapter it would violate. Routing and both refusal paths unchanged.

Worth recording, since it bears on how much the earlier rounds actually verified: **my own check of the parameter rename in `c75500f` was inconclusive.** It compared the `complete` lane *with* the flag against the `vision` lane *without* it — two paths that differ for reasons unrelated to the flag. It looked like a passing before/after test and could not have failed. Had I varied only the flag on one lane, the dead code would have surfaced four rounds earlier.

That's the third distinct instance in this PR of the same underlying mistake: a check that ran clean without actually exercising the thing it claimed to cover (`--spec` exiting 0 on wrong arithmetic, `--sample` exiting 0 on an unvalidated field, and now a before/after comparison with two variables moving). Cheap rule for next time: change one variable, and confirm the test fails when the fix is removed.

---

alirezarezvani:
Both observations checked. One closed in `5f5d84f`, one confirmed out of scope.

**Regime bands now overridable.** You called it fine as shipped, and it was, but the inconsistency is small enough to close rather than carry — and this particular band matters more than a typical default, since it decides whether `smaller-model` gets ranked last, which is the entire double-descent correction the tool exists to apply. A user whose interpolation point sits elsewhere had no way to say so.

Added `--underparameterized-max` and `--overparameterized-min`, threaded through `plan()` / `classify_regime()` as keyword args defaulting to the existing constants, and all three thresholds are now reported in `--output json` the way `training_diagnostics.py` already does.

Verified defaults preserve behavior exactly — the `--sample` fixture still reports `overparameterized / OVERFIT / 240.0` with `smaller-model` last, byte-identical to before — and that an override genuinely moves the regime (`--overparameterized-min 500` reclassifies the same model as `near-interpolation`). Exit codes unchanged: 1 action, 0 balanced, 4 leaky split.

**Duplicate `## [Unreleased]` header — confirmed pre-existing, left alone.** Checked rather than assumed:

```
count at merge-base with main : 2
count at HEAD                 : 2
```

So this branch didn't introduce it, and my entry went under the first header. Agreed it's changelog debt worth a separate cleanup — flagging it here so it doesn't get lost.

For the record on this thread: that's seven review rounds, six of which surfaced something real in this plugin's tooling — every one in code I had reported as verified. The content and rights reasoning have gone unchallenged throughout; the defects clustered entirely in the four scripts. Worth weighing against any "ready to merge" verdict, including the ones already given.

---

alirezarezvani:
Real, reproduced, and fixed in `35f5b3b`. `--goal "why overfitting still rages in large models"` exited 3 — refused as out-of-scope retrieval-augmented-generation work, exactly as you predicted.

Before fixing I checked whether the `-es` branch earns its keep at all, and it doesn't: **every keyword in the table ending in s, x, z, ch or sh is already plural or non-count** (`basics`, `features`, `foundations`, `prerequisites`, `statistics`, `speech`, `mlops`, `fairness`, `mathematics`, `time series`). Nothing needs `-es` appended, so I narrowed the auto-plural to plain `-s` rather than special-casing `rag`, and documented that irregular forms belong in `SURFACE_FORMS` spelled out — the way `fine-tuning`, `prompting` and `agentic` already are. That closes the class instead of the instance.

Verified both directions: `"rages"` no longer matches while standalone `"rag"` still does; `transformers`, `embeddings`, `autoencoders`, `agentic` still match; the goal above is now correctly *unroutable* (exit 4) rather than confidently misrouted. RLHF, LoRA fine-tuning, RAG pipelines and prompt/agent goals all still exit 3; routing unchanged across the battery.

The pattern worth naming: **this is the second round of false out-of-scope refusals from the same root cause** — approximate string matching standing in for a word list. Round 4 caught `"rag"` inside *storage*; the `-es` suffix I added to fix it reintroduced a narrower version of the same failure. The refusal path is the part of this tool a user is least able to second-guess — it states a confident conclusion and stops — so it's the wrong place to approximate. Worth remembering if these keyword tables grow.

Also noted and appreciated: you were explicit that your check was inspection rather than execution, since script execution was gated in your environment. That's useful calibration — the last three findings across reviewers all came from reading the code, not running it, and this one was reproducible in one command.

---

alirezarezvani:
Reproduced and fixed in `942e771` — and the consequence is sharper than "silently classifies via whichever branch is checked first."

With `--underparameterized-max 20 --overparameterized-min 5` and a ratio of 10, the tool reported a model **ten times overparameterized** as `underparameterized` and exited 0. That verdict ranks `smaller-model` *first* rather than last — the precise inversion of the double-descent correction this tool exists to apply. So the bad output isn't a mislabel, it's the opposite recommendation on the one axis the tool was built to get right.

Added an argparse guard rejecting `under-max >= over-min` with both values named (exit 2, the documented usage-error code). Equal bands are rejected too, since they leave `near-interpolation` unreachable. Verified: inverted and equal both exit 2; a valid override still moves the regime (`--overparameterized-min 500` → `near-interpolation`); defaults unchanged at `overparameterized / OVERFIT / 240.0` with `smaller-model` last; other exit codes still 1 / 0 / 4. Full gate suite green.

Worth recording about these two flags specifically: **they were added to close a consistency nit, and introduced this defect in doing so.** A new option is new surface, and its interaction with the options already there is part of that surface — which I didn't check when adding them. That's the second time in this PR a fix has created the next finding (the `-es` plural did the same thing to the substring fix).

Also appreciated that you re-derived the eight prior fixes independently instead of trusting the commit messages. Given that six of the nine rounds found something real in code I'd already reported as verified, re-checking rather than trusting was the right call.

---

alirezarezvani:
Checked both observations. Neither warrants a change, and here's the evidence for each rather than just a judgment call.

**1. Band rules only in comments, not `--help`.** Partly already covered, and misattributed — the regime bands and `smaller-model` demotion live in `capacity_planner.py`, not `reading_path_planner.py`. That file's `--help` does explain both bands:

```
--underparameterized-max   params-per-example below which the classical U-curve
                           applies directly (default: 1.0)
--overparameterized-min    params-per-example above which double descent makes
                           'shrink the model' unreliable (default: 10.0). These are
                           heuristic bands, not a threshold theorem — the
                           interpolation point depends on task, architecture and
                           label noise.
```

What `--help` doesn't spell out is boundary *inclusivity* (`<` vs `<=`). That's real but marginal: `--output json` echoes the exact thresholds used alongside the resulting regime, so the mapping is observable. Not worth resetting a green CI run for.

**2. `plugin.json` version 2.12.0.** The premise doesn't hold up against the tree. Counting every plugin manifest:

| version | plugins | e.g. |
|---|---|---|
| 2.9.0 | 58 | chief-ai-officer-advisor, … |
| 2.11.2 | 8 | agent-memory, memory-engineering, skill-doctor |
| 1.0.0 | **6** | agent-harness, book-to-skill, collab-proof |
| 2.10.3 | 4 | zero-hallucination-coder, roast |
| 2.12.0 | 2 | deep-learning-book, linkedin |

Repo-release-version is the dominant convention by roughly 12:1, and `2.12.0` matches the current in-flight release and the most recent additions. `book-to-skill` at `1.0.0` is one of six exceptions, not the norm. Changing mine would move it away from the majority.

**3. CI status.** `UNSTABLE` was a pending check, not a failure. Now resolved — verified against the run's own `head_sha` rather than the check listing, since a green result described a stale commit seven times earlier in this thread:

```
CI Quality Gate run #1305 → head_sha 942e771 → success
claude-review, VirusTotal, Detect changed skills → success
Tessl, Security audit → skipped by design
```

All seven checks green on the current head.

Ten rounds in, this is the first with nothing to fix — so I'm stopping here rather than making changes that don't earn their CI cycle. Still an open draft; the `marketing/linkedin` findings from round four remain for a separate PR.

---

PR #997, opened 2026-08-26
fix(linkedin): repair the policy gate's self-contradictions and six related defects

alirezarezvani:
## Summary

Addresses the `marketing/linkedin` findings raised during **#994**'s review. They were deliberately left out of that PR because the plugin was already-merged content riding along in the diff; this is the focused follow-up. Every finding was **reproduced before being fixed**, and every fix verified in both directions.

## The headline defect: two rules refused the action their own substitute recommends

| Request | Before | After |
|---|---|---|
| `export my connections list as a csv` | **REFUSE (4)** | ALLOW (0) |
| `LinkedIn's native scheduler to auto-post weekly` | **REFUSE (4)** | ALLOW (0) |

`P2-SCRAPING` told the user to run *"LinkedIn's own export of YOUR data"* while refusing the phrase for doing exactly that. `P1-AUTOMATION` refused native scheduling that `P7`'s substitute names as **the supported path**.

Rather than loosening the patterns — which would have opened a real hole — `_scan` now supports **per-rule exemptions that drop a single matched snippet, never the whole rule**. A sentence mixing an endorsed action with a prohibited one still refuses:

```
native scheduler to auto-post AND auto-connect with 500 recruiters  -> 4
export my connections list AND scrape their emails                  -> 4
export a leads database from linkedin                               -> 4   (not the self-export)
PhantomBuster + auto-DM 500                                         -> 4   (control)
write me a post about my product launch                             -> 0   (control)
```

The exemption applied is reported in the output rather than silently swallowed.

**`P3` pod rule:** the lookahead could only match noun-before-verb (`pods to join`), so the common `join a pod` was *structurally unmatchable*. Both phrasings now refuse.

## The bypassable guarantee

`outreach_message_builder.py` checked only the `--ask` field, so the same ask moved into `--reason` or `--specific-line` passed clean — making the documented *"refuses an ask in a first-touch note"* guarantee bypassable (`0` → now `3`). The note body is now scanned with a meeting-request pattern rather than the loose `ASK_RE`, which would have flagged *"your post on on-call rotations"*. Verified that innocent phrasing still passes.

## Zero-dispersion data labelled everything a breakout

With identical rates the IQR is 0, every fence collapses onto the median, and the `>= hi_fence` test fires for every post:

```
12 identical posts   before: {BREAKOUT: 12}   after: {TYPICAL: 12}
```

Varied data is unchanged (`{TYPICAL: 6, STRONG: 3, WEAK: 3}`).

## Two profile tools counted prose as signal

The auditor counted bare `to`/`from`/`x` substrings, so **"Reported to the VP of Engineering"** scored as outcome-carrying. Multipliers and from/to deltas are now shapes requiring a number.

`headline_scorer` counted bare `for `/`to `/`from `/`into `, so *"excited to share my thoughts from the conference"* scored both audience **and** outcome. **Removing them alone cost a real signal** — the sample's *"Head of Data **for** Series A/B SaaS"* names an audience through the construction, and the score dropped 93→85. So the directed-at construction is restored as a shape that must land on an actual audience noun: sample back to **93**, weak sample unchanged at **19**, prose false positives gone.

## Typed exit codes instead of tracebacks

All **eleven** scripts crashed with a `FileNotFoundError` traceback and exit 1 on a mistyped `--input`; each now exits 2 with a message. The two analytics scripts died with `AttributeError` on a bare list of scalars; both now exit 4 naming the problem.

Worth noting: catching `json.JSONDecodeError` does **not** catch `ValueError` — the subclass relation runs the other way — so the analyzer's `except` clause was widened rather than left to trade one traceback for another.

## Two smaller corrections

- `cadence_planner` reported `FITS`/exit 0 while handing back a week containing **zero posts**; now `NO_POSTS_AFFORDABLE`/exit 3 with a blocking finding and an updated exit table. Normal budgets unaffected, below-floor still exits 2.
- `pattern_miner`'s multiple-comparisons note now states plainly that it is an **accounting** of expected false positives, not an applied Bonferroni or BH correction, so the number cannot be read as a stronger guarantee than it is.

## Checklist

- [x] **Target branch is `dev`** (not `main`)
- [x] Scripts run with `--help` without errors
- [x] No hardcoded API keys, tokens, or secrets — no network calls, no LinkedIn API access
- [x] No vendor-locked dependencies — standard library only

## Type of Change

- [ ] New skill
- [x] Improvement to existing skill
- [x] Bug fix
- [ ] Documentation
- [ ] Infrastructure / CI

## Testing

Blocking gates, all green: `compileall`, `check_paths`, `check_frontmatter`, `check_dual_publish`, `check_model_freshness`, `smoke_scripts` (**696 passed / 0 failed**), `derive_counters --check`, `check_skill_names`, `check_plugin_json`.

Behavioural verification is the table above — each row run against `HEAD` and against the working tree. Sample outputs pinned: headline 93 (was 93 before the over-broad edit), weak headline 19, profile auditor 37/WEAK, cadence `--sample` exit 3 for its own pre-existing reason.

The advisory JSON-output gate still reports its **8 pre-existing `agent-launcher` failures; none is in this plugin** — untouched here deliberately, and worth a separate issue.

---
_Generated by [Claude Code](https://claude.ai/code/session_01BswsZp5zrJWFAGU6KWNA1s)_

### DIFF
--- a/marketing/linkedin/skills/linkedin-analytics/scripts/pattern_miner.py
+++ b/marketing/linkedin/skills/linkedin-analytics/scripts/pattern_miner.py
@@ -276,7 +276,10 @@ def _read_input(path: str) -> str:
     try:
         with open(path, encoding="utf-8") as handle:
             return handle.read()
-    except OSError as err:
+    except (OSError, UnicodeDecodeError) as err:
+        # UnicodeDecodeError is a ValueError subclass, not an OSError, so a file that
+        # exists but is not valid UTF-8 escaped the OSError catch as a traceback -
+        # the same failure mode this helper exists to prevent for a bad path.
         print(f"cannot read --input {path}: {err}", file=sys.stderr)
         raise SystemExit(2)
 
--- a/marketing/linkedin/skills/linkedin-analytics/scripts/pattern_miner.py
+++ b/marketing/linkedin/skills/linkedin-analytics/scripts/pattern_miner.py
@@ -225,6 +225,9 @@ def mine(rows: list, attributes: list) -> dict:
         "multiple_comparisons_note": (
             f"{len(tested)} candidate(s) reached the test at alpha {ALPHA}. On noise alone you "
             f"would expect about {expected_false} to pass. {len(supported)} did. "
+            "This is an accounting of that expectation, not an applied correction: no "
+            "Bonferroni or Benjamini-Hochberg adjustment is made to the per-candidate "
+            "threshold, so a single passing candidate is weak evidence on its own. "
             + ("Treat these as hypotheses to test deliberately, not as conclusions."
                if len(supported) <= max(1, expected_false)
                else "More passed than chance predicts, which is mild evidence something real is "
@@ -262,6 +265,22 @@ def render_human(r: dict) -> str:
     return "\n".join(lines)
 
 
+def _read_input(path: str) -> str:
+    """Read --input, turning an unreadable path into a usage error, not a traceback.
+
+    Every script in this plugin caught a JSON decode error but let OSError escape, so
+    a mistyped path exited 1 with a FileNotFoundError stack instead of a typed code.
+    """
+    if path == "-":
+        return sys.stdin.read()
+    try:
+        with open(path, encoding="utf-8") as handle:
+            return handle.read()
+    except OSError as err:
+        print(f"cannot read --input {path}: {err}", file=sys.stderr)
+        raise SystemExit(2)
+
+
 def main() -> int:
     ap = argparse.ArgumentParser(
         description="Test candidate LinkedIn patterns against a permutation null "
@@ -278,7 +297,7 @@ def main() -> int:
     if args.sample:
         rows = SAMPLE
     elif args.input:
-        raw = sys.stdin.read() if args.input == "-" else open(args.input, encoding="utf-8").read()
+        raw = _read_input(args.input)
         try:
             rows = ([dict(r) for r in csv.DictReader(io.StringIO(raw))] if args.csv
                     else json.loads(raw))
@@ -287,6 +306,10 @@ def main() -> int:
             return 4
         if isinstance(rows, dict):
             rows = rows.get("posts") or rows.get("rows") or []
+        if not isinstance(rows, list) or any(not isinstance(r, dict) for r in rows):
+            print("ERROR: input must be a list of post objects (a bare list of values "
+                  "cannot be mined).", file=sys.stderr)
+            return 4
     else:
         ap.error("--input or --sample is required")
 
--- a/marketing/linkedin/skills/linkedin-analytics/scripts/post_performance_analyzer.py
+++ b/marketing/linkedin/skills/linkedin-analytics/scripts/post_performance_analyzer.py
@@ -95,6 +95,16 @@ def load_rows(raw: str, as_csv: bool) -> list:
     data = json.loads(raw)
     if isinstance(data, dict):
         data = data.get("posts") or data.get("rows") or []
+    if not isinstance(data, list):
+        raise ValueError(f"expected a list of post objects, got {type(data).__name__}")
+    # A bare list of scalars ([1,2,3]) used to reach row.get() and die with an
+    # AttributeError traceback instead of a typed exit code.
+    bad = next((r for r in data if not isinstance(r, dict)), None)
+    if bad is not None:
+        raise ValueError(
+            f"every row must be an object with an 'impressions' field; found a "
+            f"{type(bad).__name__} ({bad!r:.40})"
+        )
     return data
 
 
@@ -142,9 +152,17 @@ def analyse(rows: list) -> dict:
     q1, q3 = percentile(ers, 25), percentile(ers, 75)
     iqr = q3 - q1
     hi_fence, lo_fence = q3 + 1.5 * iqr, q1 - 1.5 * iqr
+    # With no dispersion (identical or heavily rounded rates) every fence collapses
+    # onto the median, and the >= hi_fence test below would label every post
+    # BREAKOUT - the opposite of describing the data honestly. There is no spread
+    # to rank, so nothing is an outlier.
+    degenerate = iqr == 0
 
     for p in clean:
         e = p["engagement_rate"]
+        if degenerate:
+            p["band"] = "TYPICAL"
+            continue
         if e >= hi_fence:
             p["band"] = "BREAKOUT"
         elif e >= q3:
@@ -219,6 +237,22 @@ def render_human(r: dict) -> str:
     return "\n".join(lines)
 
 
+def _read_input(path: str) -> str:
+    """Read --input, turning an unreadable path into a usage error, not a traceback.
+
+    Every script in this plugin caught a JSON decode error but let OSError escape, so
+    a mistyped path exited 1 with a FileNotFoundError stack instead of a typed code.
+    """
+    if path == "-":
+        return sys.stdin.read()
+    try:
+        with open(path, encoding="utf-8") as handle:
+            return handle.read()
+    except OSError as err:
+        print(f"cannot read --input {path}: {err}", file=sys.stderr)
+        raise SystemExit(2)
+
+
 def main() -> int:
     ap = argparse.ArgumentParser(
         description="Describe your own LinkedIn post export honestly "
@@ -236,10 +270,12 @@ def main() -> int:
     if args.sample:
         rows = SAMPLE
     elif args.input:
-        raw = sys.stdin.read() if args.input == "-" else open(args.input, encoding="utf-8").read()
+        raw = _read_input(args.input)
         try:
             rows = load_rows(raw, args.csv)
-        except (json.JSONDecodeError, csv.Error) as exc:
+        except (ValueError, csv.Error) as exc:
+            # ValueError covers json.JSONDecodeError (its subclass) and the explicit
+            # row-shape errors load_rows raises.
             print(f"ERROR: could not parse input: {exc}", file=sys.stderr)
             return 4
     else:
--- a/marketing/linkedin/skills/linkedin-analytics/scripts/post_performance_analyzer.py
+++ b/marketing/linkedin/skills/linkedin-analytics/scripts/post_performance_analyzer.py
@@ -248,7 +248,10 @@ def _read_input(path: str) -> str:
     try:
         with open(path, encoding="utf-8") as handle:
             return handle.read()
-    except OSError as err:
+    except (OSError, UnicodeDecodeError) as err:
+        # UnicodeDecodeError is a ValueError subclass, not an OSError, so a file that
+        # exists but is not valid UTF-8 escaped the OSError catch as a traceback -
+        # the same failure mode this helper exists to prevent for a bad path.
         print(f"cannot read --input {path}: {err}", file=sys.stderr)
         raise SystemExit(2)
 
--- a/marketing/linkedin/skills/linkedin-analytics/scripts/post_performance_analyzer.py
+++ b/marketing/linkedin/skills/linkedin-analytics/scripts/post_performance_analyzer.py
@@ -154,14 +154,19 @@ def analyse(rows: list) -> dict:
     hi_fence, lo_fence = q3 + 1.5 * iqr, q1 - 1.5 * iqr
     # With no dispersion (identical or heavily rounded rates) every fence collapses
     # onto the median, and the >= hi_fence test below would label every post
-    # BREAKOUT - the opposite of describing the data honestly. There is no spread
-    # to rank, so nothing is an outlier.
+    # BREAKOUT - the opposite of describing the data honestly.
+    #
+    # But iqr == 0 only means the middle 50% is flat, which is a realistic shape for
+    # a real export: many similar posts plus a couple of viral ones. Collapsing
+    # everything to TYPICAL there hides genuine breakouts, trading one honesty
+    # problem for another. So when the centre is degenerate, rank against the median
+    # itself: equal to it is TYPICAL, above it is a real standout, below it is weak.
     degenerate = iqr == 0
 
     for p in clean:
         e = p["engagement_rate"]
         if degenerate:
-            p["band"] = "TYPICAL"
+            p["band"] = "TYPICAL" if e == med else ("BREAKOUT" if e > med else "WEAK")
             continue
         if e >= hi_fence:
             p["band"] = "BREAKOUT"
--- a/marketing/linkedin/skills/linkedin-content/scripts/post_linter.py
+++ b/marketing/linkedin/skills/linkedin-content/scripts/post_linter.py
@@ -306,7 +306,10 @@ def _read_input(path: str) -> str:
     try:
         with open(path, encoding="utf-8") as handle:
             return handle.read()
-    except OSError as err:
+    except (OSError, UnicodeDecodeError) as err:
+        # UnicodeDecodeError is a ValueError subclass, not an OSError, so a file that
+        # exists but is not valid UTF-8 escaped the OSError catch as a traceback -
+        # the same failure mode this helper exists to prevent for a bad path.
         print(f"cannot read --input {path}: {err}", file=sys.stderr)
         raise SystemExit(2)
 
--- a/marketing/linkedin/skills/linkedin-content/scripts/post_linter.py
+++ b/marketing/linkedin/skills/linkedin-content/scripts/post_linter.py
@@ -295,6 +295,22 @@ def render_human(r: dict) -> str:
     return "\n".join(lines)
 
 
+def _read_input(path: str) -> str:
+    """Read --input, turning an unreadable path into a usage error, not a traceback.
+
+    Every script in this plugin caught a JSON decode error but let OSError escape, so
+    a mistyped path exited 1 with a FileNotFoundError stack instead of a typed code.
+    """
+    if path == "-":
+        return sys.stdin.read()
+    try:
+        with open(path, encoding="utf-8") as handle:
+            return handle.read()
+    except OSError as err:
+        print(f"cannot read --input {path}: {err}", file=sys.stderr)
+        raise SystemExit(2)
+
+
 def main() -> int:
     ap = argparse.ArgumentParser(
         description="Lint a LinkedIn post 0-100 (SHIP=0 / REVISE=2 / REWRITE=3).")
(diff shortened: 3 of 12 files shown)


alirezarezvani:
All three findings reproduced and fixed in `4dbb60f`. The first one was a false claim in my own PR description, which is worth stating plainly.

**1. Exemption transparency — you were right, and the description overstated it.** `_scan` attached the excused list to a *hit*, and a hit only exists when the rule still has an unexempted match. So when **every** match was exempted — precisely the two headline examples — no hit existed and the trace was discarded:

```
before:  verdict=ALLOW   exemptions in output: none
after:   verdict=ALLOW   exemptions_applied: [('P2-SCRAPING', 'export my connections list as a csv')]
```

Exemptions are now collected regardless of verdict, so the description's claim is true as written rather than true-only-for-the-mixed-case.

**2. `render_human` never printed them.** Correct — JSON-only, in the mode most likely to be read interactively. It now prints a `Recognized but exempt` block.

**3. P1 exemption narrower than the bug class.** Confirmed: `automate posting with LinkedIn's native scheduler` matched `\bautomat(e|ed|ing|ion)\b`, produced the snippet `automate`, missed `excuses`, and still refused — one surviving instance of the exact self-contradiction this PR exists to fix. It wasn't deliberate scoping so much as an incomplete fix.

You correctly anticipated why it couldn't just be widened: the bare token `automate` is far too generic to exempt on the scheduler signal alone. Exemptions now support an optional **positional `context`** that must appear just after the matched word:

```
automate posting with LinkedIn's native scheduler        -> 0  (allowed)
use LinkedIn's native scheduler and automate my DMs      -> 4  (still refuses)
```

**4. `_read_input` duplication — deliberate, now stated.** CLAUDE.md lists cross-skill dependencies as an anti-pattern and these scripts live in six different skills, so each stays independently extractable. Your point about eleven synchronized edits is the real cost of that trade, and it's the trade the repo has chosen.

---

One verification note, since it bears on how much my earlier evidence was worth. While re-checking the ask-detection fix I extracted the "before" baseline from `HEAD` — but HEAD was already my own fix commit, so I compared the fix against itself and got a result implying the bypass had never existed. The true baseline is `origin/dev`:

```
origin/dev (pre-fix):  exit=0  premature-ask=False     <- the bypass
mine (fixed):          exit=3  premature-ask=True
innocent 'on-call':    exit=0  before and after        <- no false positive
```

That's the second time in these two PRs that a check of mine ran clean without exercising what it claimed to. Full regression battery re-run — all eleven policy-gate cases hold including every no-bypass control — and all blocking gates green (`smoke_scripts` 696 passed).

---

alirezarezvani:
Both reproduced and fixed in `97bc4ac`. The de-dup bug was a good catch — it was sitting in the exact code the previous commit added to make exemptions trustworthy.

**1. De-dup never fired.** Exactly as you traced: `excused` holds dicts, so `snippet in excused` compares a `str` against dicts and is always `False`, while the neighbouring `matched` list holds plain strings and de-dupes correctly.

```
"auto-post Monday with the native scheduler, then auto-post again Thursday"
  before:  exemptions_applied: ['auto-post', 'auto-post']
  after:   exemptions_applied: ['auto-post']
```

No verdict changed, but you put the real cost precisely: the point of surfacing `exemptions_applied` was to make that list trustworthy, and an inflated list is not. Snippet strings now live in a `set()` alongside the dicts. Verified the obvious over-correction didn't happen either — two **different** exempt phrases are still both reported:

```
"export my connections list as a csv, and auto-post with LinkedIn's native scheduler"
  -> [('P1-AUTOMATION', 'auto-post'), ('P2-SCRAPING', 'export my connections list as a csv')]
```

**2. `into` dropped with no replacement.** Confirmed asymmetric — `from ` kept three idiomatic stand-ins and `into ` got none. Restored as the *construction* rather than the preposition, reusing the shapes mechanism already in the file:

```
"took the team from manual QA into automated coverage"   outcome 20
"putting budget into growth"                             outcome 12  (via "grow" only)
"excited to share my thoughts from the conference"       outcome  0
```

Ordinary prose still scores nothing, which was the reason the bare marker came out.

**3. `MEETING_ASK_RE` gerunds — not changed, and I agree with your framing.** "arranging a meeting" slips through. It's a content linter, not a security boundary, and widening toward gerunds is exactly how the false positives it was written to avoid ("your post on **on-call** rotations") creep back in. Noting it rather than trading a real false-positive risk for a theoretical false negative.

Regression re-run: all 12 policy-gate cases hold, and the sample outputs are pinned unchanged — headline 93/SHIP (audience 20, outcome 20), weak 19, profile auditor 37/WEAK, analyzer ANALYSED with 6 TYPICAL / 3 STRONG / 3 WEAK. All blocking gates green, `smoke_scripts` 696 passed.

---

alirezarezvani:
You were right, and this was the serious one. Reproduced against `origin/dev` before touching anything:

```
"I want to auto-post daily to 50 LinkedIn groups without permission, promoting my
 course. LinkedIn's native scheduler is neat, right?"

   pre-PR:  exit 4 (REFUSE)        with my exemption:  exit 0 (ALLOW)
```

My change turned a refusal into an allow. That is the one direction this gate must never move by accident, and the mechanism I added to fix a self-contradiction did it. Fixed in `81c2c81`.

**Root cause exactly as you traced it.** `signal` was searched across the whole input, so an endorsement in one sentence retroactively excused an unrelated match in another. `signal` is now matched only within the sentence containing the match. The trade is stated in the docstring rather than hidden: a legitimate two-sentence phrasing ("I use LinkedIn's native scheduler. I auto-post weekly.") now refuses and the user can rephrase — the alternative cost was a false ALLOW they cannot detect at all.

**Then the adversarial battery found a hole your example didn't reach.** With the endorsement in the *same* sentence, `auto-post daily to 50 groups without permission using LinkedIn's native scheduler` was still allowed. The exemption isn't wrong — native scheduling *is* the supported path P7 names. The real gap was that **no rule covered unsolicited bulk posting**; pre-PR the gate refused that text only incidentally, by refusing every automation word. So P5 is broadened from messaging to messaging-or-posting, and now refuses for the actual reason.

That rule also closes a gap the gate always had: `schedule my posts to 40 communities without permission` has no automation word and was **allowed pre-PR (exit 0)**. It now refuses.

```
ADVERSARIAL CASE                                    PRE-PR   NOW
spam 50 groups + aside in NEXT sentence                  4     4
spam 50 groups + endorsement SAME sentence               4     4
bulk group posting, no automation word                   0     4   <- gap that predates this PR
scheduler aside + auto-connect                           4     4
scheduler + auto-dm (semicolon)                          4     4
own-data aside + leads DB export                         4     4
LEGIT self-export                                        4     0
LEGIT native auto-post                                   4     0
LEGIT automate posting                                   4     0
LEGIT small-scale group posting ("3 groups I belong to") 0     0
control: phantombuster                                   4     4
control: benign                                          0     0
```

**Finding 2 confirmed and fixed too.** `UnicodeDecodeError` is a `ValueError` subclass, not an `OSError`, so an existing non-UTF-8 file escaped as a traceback — the same failure mode the helper exists to prevent. Fixed in all 11 scripts; it now exits 2 with a message.

Other tools unchanged (headline 93/SHIP, profile 37/WEAK, analyzer ANALYSED 6/3/3, cadence NO_POSTS_AFFORDABLE/3), all blocking gates green.

Worth stating plainly for the record: **this is the fourth time in this PR pair that a fix of mine contained the next defect**, and the only one that weakened a safety gate. My own "no bypass" controls missed it because every one of them relied on a *second* pattern tripping to save the verdict — none tested the case where the only match is the exempted one. A control that cannot fail is not a control.

---

alirezarezvani:
## Round 4 — three findings fixed, pushed as `c4f1d33`

### 1. De-duplication decided exemptions (High)

The gate de-duplicated matches by snippet **text** before deciding whether an exemption applied, so the second use of a generic word inherited the first one's fate:

```
automate posting via linkedin's native scheduler, and also
automate direct messaging to my whole network              before: 0 ALLOW   after: 4 REFUSE
```

The first `automate` was excused (native scheduling is endorsed), the second was skipped as a duplicate — so the mass-DM half was never evaluated. Each occurrence is now judged at its own position; de-duplication applies to the display lists only.

This one is mine: the exemption mechanism that introduced it was added in this PR.

### 2. A flat centre hid real breakouts (Medium)

The previous round's fix labelled every post `TYPICAL` when `iqr == 0`. But a flat middle 50% is a realistic export shape — many similar posts plus a couple of viral ones — so that traded one honesty problem for another. A degenerate centre now ranks against the median:

```
12 identical posts            -> {TYPICAL: 12}
10 flat + 2 viral             -> {TYPICAL: 10, BREAKOUT: 2}
10 flat + 2 duds              -> {TYPICAL: 10, WEAK: 2}
normal spread (regression)    -> {TYPICAL: 6, WEAK: 3, STRONG: 2, BREAKOUT: 1}
```

### 3. `communities` listed twice in P5 (Nit)

Second alternative is now `forums?`, which the rule could not previously match. `post my article into 30 forums automatically` → `0` before, `4` now.

---

### Verification

An 18-case adversarial battery (12 must-refuse, 6 must-allow): **18/18 correct**.

Refusals were **attributed to the specific rule** rather than assumed — the three automation cases refuse on `P1-AUTOMATION` with the same word in both `matched` and `exempted`, which is the fix firing, not another rule covering for it. Running the same cases against the pre-fix file confirms 3 of 4 returned `ALLOW` before this change (the reversed-order case already refused, because there the prohibited occurrence came first — order mattered, which is the bug restated).

That control matters here because in the previous rounds every "no bypass" check I ran depended on a *second* pattern tripping to save the verdict, so none of them could have caught this class. A control that cannot fail is not a control.

Blocking gates green: `compileall`, `check_paths`, `check_frontmatter`, `check_dual_publish`, `check_model_freshness`, `check_skill_names`, `check_plugin_json --all`, `derive_counters --check`, `smoke_scripts` **696 passed / 0 failed**. Pinned sample outputs unchanged: headline 93/SHIP, weak 19, profile auditor 37/WEAK, analyzer `ANALYSED` 6 TYPICAL / 3 STRONG / 3 WEAK.

---

alirezarezvani:
## Round 5 — both findings confirmed and fixed, pushed as `8a16de8`

Both reproduced before touching anything. Ran for real, since the review flagged that it couldn't execute Python.

### 1. `context` window bled across the clause boundary — confirmed

The repro is exact:

```
automate connects and automate posting via native scheduler.    ->  0 ALLOW
automate connects. and automate posting via native scheduler.   ->  4 REFUSE
```

One period apart. The second clause's `posting` was landing inside the first `automate`'s 40-character window, so the endorsed action excused the prohibited one.

The trace was right about the mechanism and about it being the same class as the round-3 `signal` fix, still open on the `context` path. The comment there claimed the word had to appear *"immediately after the matched word"* — a fixed character count was never that.

`_clause_after` now cuts the window at the first clause boundary (comma, sentence end, or a coordinating conjunction: `and`/`but`/`then`/`also`/`plus`/`while`/`or`/`as well as`). I kept the character cap as a ceiling rather than replacing it — a clause boundary alone leaves one long unpunctuated clause able to drift, e.g. `automate connecting with recruiters for my posting campaign`. Both bounds apply, whichever is tighter; that case refuses.

Verified in both directions:

| input | before | after |
|---|---|---|
| `automate connects and automate posting via native scheduler.` | 0 ALLOW | **4 REFUSE** |
| `automate connecting and automate posting…` | 0 ALLOW | **4 REFUSE** |
| `automate outreach and automate posting…` | 3 CONSTRAIN | **4 REFUSE** |
| `automate connecting with recruiters for my posting campaign…` | — | **4 REFUSE** |
| `automate posting via linkedin's native scheduler` | 0 ALLOW | 0 ALLOW |
| `automate publishing my content with the native scheduler` | 0 ALLOW | 0 ALLOW |
| round-4 case: exempt clause then mass-DM clause | 4 REFUSE | 4 REFUSE |

### 2. `led` / `cut` substring false positives — confirmed

`w in low` credited five duty bullets:

```
Scheduled onboarding for new hires   -> ['led']        Installed the new CI runner  -> ['led']
Handled escalations from support     -> ['led']        Recalled prior incidents     -> ['led']
Executed the Q3 roadmap              -> ['cut']
```

Right that this is the defect removing bare `to`/`from`/`x` was meant to end, and that these two survived it. Now matched on word boundaries via a compiled alternation. `%` leaves the tuple entirely — `OUTCOME_SHAPES` already carries the stricter `\d+\s*%`, so a stray percent sign no longer counts as a result on its own.

After: 7 duty bullets uncredited, 6 genuine outcome bullets (`Cut infra spend 30%`, `Led the platform migration`, `Grew ARR from 2M to 9M`, …) still credited.

### On the nit

Agreed, and no action taken — `_read_input` duplication stays.

---

**Verification:** 18-case adversarial battery 18/18. Blocking gates green — `compileall`, `check_paths`, `check_frontmatter`, `check_dual_publish`, `check_model_freshness`, `check_skill_names`, `check_plugin_json --all`, `derive_counters --check`, `smoke_scripts` **696 passed / 0 failed**. Pinned samples unchanged: headline 93/SHIP, profile auditor 37/WEAK, analyzer 6 TYPICAL / 3 STRONG / 3 WEAK.

---

alirezarezvani:
## Round 6 — finding confirmed and fixed, pushed as `8117640`

### The `ASK_RE` false negative is real

The diagnosis was right, and the symmetry is the interesting part: the connection branch was fixed earlier in this PR because a loose whole-note match **raised a false finding**; the `else` branch had the same defect running the other way, where it **suppresses a real one**.

Reproduced on `dm`:

| note | `pitch-without-ask` |
|---|---|
| pitch language, no ask, neutral text | fires |
| …plus `your post on on-call rotations` | **silently lost** |
| …plus `i saw the chat transcript feature` | **silently lost** |
| …plus `watched your demo video` | **silently lost** |

An ask is now the `--ask` field or meeting-request framing in the note — the same test the connection branch already uses. Verified the three suppressed cases fire, and three controls stay quiet: a real ask in `--ask`, a meeting-shaped ask in the note body, and a note with no pitch at all.

One thing the fix surfaced: it left `ASK_RE` referenced nowhere. I removed it rather than leave a dead pattern, and moved its rationale into the comment on `MEETING_ASK_RE`, which now records both directions the defect ran in — otherwise the next person reads a comment explaining a constraint on a regex that no longer exists.

**A note on my own process here:** my first repro of this used `message_type` as the input key. The script reads `type`, so the field was ignored, the default `connection` was used, and I traced the wrong branch entirely — the output showed `pitch-on-first-touch`, a connection-branch finding, which is what caught it. Worth stating because the failure mode is exactly the one that has bitten this PR repeatedly: a probe that runs cleanly while exercising something other than what it claims.

### `cadence_planner` help text

Fixed — the argparse description now mentions `NO_POSTS_AFFORDABLE` alongside over-budget on exit 3.

### The two nits I did not act on

- **`_read_input` duplication.** Agreed, left as-is; it's the accepted cost of the self-contained-package convention.
- **Two shapes of bare-list validation** (`pattern_miner` returns `4` inline, `post_performance_analyzer` raises `ValueError` for `main()` to catch). Both are correct. I left them: churning one to match the other adds diff without changing behaviour, and neither is the shape a third script would obviously copy. Flagging that I made that call rather than silently skipping it.

---

**Verification:** adversarial battery 18/18; blocking gates green — `compileall`, `check_paths`, `check_frontmatter`, `check_dual_publish`, `check_model_freshness`, `check_skill_names`, `check_plugin_json --all`, `derive_counters --check`, `smoke_scripts` **696 passed / 0 failed**. Pinned samples unchanged: headline 93/SHIP, profile auditor 37/WEAK, analyzer 6 TYPICAL / 3 STRONG / 3 WEAK, outreach `PASS` with no findings.

---

alirezarezvani:
## CI on `8117640` is failing at workflow startup — not from this diff

Flagging this rather than letting it sit, because it is not visible from the checks list: the PR does not look red, it looks *empty*.

### What is actually happening

| head | workflows created | outcome |
|---|---|---|
| `c4f1d33` (round 4) | 4 of 4 | all success |
| `8a16de8` (round 5) | 2 of 4 | success — **CI Quality Gate + VirusTotal never created** |
| `8117640` (round 6) | 2 of 4 | **both `startup_failure`** — CI Quality Gate + Claude Code Review never created |

`startup_failure` is a workflow-level failure: it happens before any step runs, so no code in this diff executes. The two that failed are **VirusTotal Security Scan** and **Skill Quality Review (Tessl)** — neither reads the Python files this PR touches, and this PR touches nothing under `.github/workflows/`.

### Why this is not this PR's failure

`dev` at `ae035f5a` — this PR's own base — ran **all five** workflows green at 13:52Z today (CI Quality Gate, Claude Code Review, VirusTotal, Skill Security Audit, Tessl). The degradation is time-correlated, not diff-correlated: it starts around 15:00Z and worsens across two pushes whose diffs are ordinary Python edits. Workflows progressively stopped being *created at all*, which no repository content can cause.

That pattern points at the Actions account/runner level — quota, spending limit, or runner availability — and it is outside what I can fix from a PR branch.

I have spent my one re-run on `32984413034` (queued 15:56Z) to confirm it reproduces. I'll report the result rather than predict it.

### The part that matters for merging

**`CI Quality Gate` has not run since `c4f1d33`.** That is the blocking gate — `check_plugin_json`, `derive_counters --check`, `smoke_scripts`, the path and frontmatter checks. The last two commits have not been validated by it.

I ran that gate's full contents locally against the current head before each push, and all of it passes:

```
compileall ................ OK          check_skill_names ......... exit 0
check_paths ............... exit 0      check_plugin_json --all ... exit 0
check_frontmatter ......... exit 0      derive_counters --check ... exit 0
check_dual_publish ........ exit 0      smoke_scripts ..... 696 passed / 0 failed
check_model_freshness ..... exit 0      adversarial battery ....... 18/18
```

That is evidence, not a substitute — a local run is not the gate. **This PR should not merge until CI Quality Gate has actually run green on the head commit.** If the Actions problem is a spending or quota limit, that needs an owner action; nothing I can push will clear it.

---

alirezarezvani:
### Re-run result: confirmed — `startup_failure` reproduces

Reporting the outcome I said I'd report rather than predict.

Run [`32984413034`](https://github.com/alirezarezvani/claude-skills/actions/runs/32984413034) re-ran on the same head `8117640` and failed at startup a second time. It also sat in `queued` for ~40 minutes before failing, and check runs on the head currently return **0**.

That closes the diagnosis in the comment above: an identical failure on one re-run, on a workflow this diff does not touch, while the base branch ran all five workflows green earlier today. This is not this PR's failure, and my one permitted re-run is now spent.

`CI Quality Gate` has still never been created for this head. Nothing further is actionable from the branch side — I'm not pushing an empty commit or reopening the PR to kick CI, since neither is a legitimate way to trigger it and neither would help if the cause is an Actions quota or spending limit.

**Still standing: this should not merge until `CI Quality Gate` runs green on the head commit.** Clearing the Actions problem needs an owner action. I'll keep watching and will report if it recovers — or, if the gate runs and is genuinely red on this diff, that one is mine and I'll fix it.

---

PR #1003, opened 2026-08-31
feat(engineering): add a verification stage and a second provider to hivemind

Hanishchow:
## Summary
- **What:** adds a verification stage and a second provider to `engineering/hivemind`, plus a `--json` fix in the worker script.
- **Why:** both are weaknesses I hit while actually running the skill after #979 merged.

## The verification stage

Every worker claim had to be re-checked by hand, which cost more than the delegation saved. `scripts/oc-verify.mjs` sends the claims to a verifier that re-derives them from the same files and tries to refute them. It returns `{confirmed, refuted, unsupported, unverified, needs_review[]}` — you read `needs_review` only; confirmed claims come back as a count rather than prose.

Four rules keep it from becoming a rubber stamp, and they're documented in SKILL.md:

- **The verifier runs on a different model from the worker.** Same model means the same blind spots.
- It sees the **files**, never the worker's reasoning.
- It **defaults to REFUTED when uncertain**, because a false CONFIRMED gets acted on unchecked.
- A verifier returning fewer verdicts than claims marks the remainder `unverified`, not passed.

## The second provider

opencode's free tier is not dependable — models return provider 404s and empty bodies. `scripts/or-worker.mjs` speaks the identical one-JSON-line contract over OpenRouter, so `oc-verify`, `oc-status` and `oc-aggregate` work against it unchanged. Key comes from `OPENROUTER_API_KEY` or `~/.claude/.openrouter_key`, is never logged, and never appears in output.

SKILL.md now includes a measured table of which models actually work: six free-model attempts failed for five different reasons, while the same free model answered a plain-text question correctly at zero cost. The rule that falls out is capability, not availability — structured output needs a paid model, prose does not.

## The `--json` fix

Workers bury their JSON in fenced blocks after a paragraph of prose. `oc-worker.mjs --json` now extracts the value and gives one corrective retry, instead of every caller writing the same regex. Unit-tested against fenced, prose-prefixed, bare-array, bare-object and fence-without-language shapes.

## Checklist
- [x] Targets `dev`
- [x] Frontmatter unchanged: `name` + `description` only
- [x] 239 lines, under the 500 limit
- [x] Anti-patterns and Cross-references sections preserved from the merged version
- [x] `assets/commands/` layout untouched — additive only, nothing removed
- [x] All three scripts pass `node --check`

Scripts remain `.mjs` rather than stdlib Python, as in #979. Happy to adjust if that's now a problem.


Hanishchow:
CI status and one thing I need your call on.

**Fixed after opening:**

- **Security audit → SUCCESS.** Overwriting `oc-worker.mjs` with my copy dropped the `// auditor:ignore-line` note added upstream after #979, which reintroduced a CRITICAL `CMD-INJECT`. Restored verbatim. `oc-verify.mjs` imports `child_process` for the same reason (it invokes `oc-worker.mjs`) so it carries its own justification rather than borrowing yours. Verified locally with `skill_security_auditor.py --strict`: PASS, 0 critical.
- **Frontmatter gate G10 → passing.** The new agent file had no `name`. The existing agents were namespaced after #979 merged, so it's `hive-verifier` now. `check_frontmatter.py`: 0 errors, same single warning your other agent files produce.

**Still red: G3 (counter derivation), and I don't think I should fix it.**

Adding one agent file takes the derived `agents` count 118 → 119. I bumped the two places a contributor can reasonably touch:

- `CLAUDE.md` "Current Scope" line
- the README `Agents-118-blue` badge

But `derive_counters.py` also validates the same claim in:

- `.claude-plugin/marketplace.json`
- `.codex-plugin/plugin.json`

CONTRIBUTING says not to modify `marketplace.json` or `.codex/`, and those look generated by your sync workflows — so editing them would be both against the rules and pointless once sync runs. As it stands, **any external PR that adds an agent can't pass G3 without breaking CONTRIBUTING.**

Three ways I'm happy to go:

1. You regenerate the derived counters after merge, and G3 is treated as expected-red here.
2. You'd rather I bump all four anyway — say so and I'll push it.
3. **Drop the agent file entirely.** `oc-verify.mjs` works with any agent id via `--model`; `verifier.md` is a convenience persona. Removing it keeps the counters untouched and the verification layer intact. This is my recommendation if you want the simplest merge.

**VirusTotal** is failing the same way it did on #979 — fork PRs don't get the API secret. **Tessl** I have no visibility into.

Happy to take whichever path is least work for you.

PR #1010, opened 2026-09-04
fix(gate): catch fabricated delta-style counter claims

### LINKED ISSUE #995: derive_counters.py --check gates absolute totals but not delta claims, and never reads CHANGELOG.md
## Summary

`scripts/derive_counters.py --check` is a strong gate on **absolute totals in five named files**, but two whole classes of counter prose sit outside it. A wrong counter can be written into `CHANGELOG.md` or a `CLAUDE.md` release note and pass CI green, every time.

This surfaced concretely in #993: the changelog and the `CLAUDE.md` release note both claimed `commands 147 → 150` for a plugin that adds exactly **one** command. The gate passed. It was caught by a human/bot reviewer reading the prose, not by CI.

## Reproduction

Appending a fabricated delta to `CHANGELOG.md` and running the gate on an otherwise clean, correct tree:

```
$ printf '\nCounters: skills 999 → 1234; commands 12 → 999.\n' >> CHANGELOG.md
$ python3 scripts/derive_counters.py --check
Counter check passed: README.md, CLAUDE.md, marketplace.json, mkdocs.yml, .codex-plugin match derived values.
```

Green, with `skills 999 → 1234` sitting in the changelog.

## The two gaps

**1. File coverage.** The gate reads `README.md`, root `CLAUDE.md` (only its `Current Scope` / `Status:` lines), `marketplace.json` `metadata.description`, `mkdocs.yml` `site_description`, and `.codex-plugin/plugin.json` descriptions. `CHANGELOG.md` is not read at all, and `CLAUDE.md`'s release-note sections — which carry a counter line per release — are outside the two checked lines.

**2. Claim shape.** `CLAIM_PATTERNS` recognises seven fixed phrasings of an *absolute total*:

```python
"skills":  r"(\d+)\s+production-ready (?:Claude Code )?skills"
"commands": r"(\d+)\s+slash commands"
...
```

Nothing matches a **delta** of the form `X → Y`, in any file — including the five that *are* checked. Since release notes and changelog entries state deltas rather than totals by convention, the dominant form of counter prose in this repo is the one form the gate cannot see.

Token-budget figures (`2,101 / 4,000 tokens`) are outside the vocabulary too, and the comma grouping would defeat `(\d+)` even if a pattern existed. #993 shipped `CLAUDE.md` claiming `2,066` on one line and `2,101` twelve lines later, about the same file.

## Why deltas are the higher-risk form

An absolute total is wrong only if it is wrong. A delta is a *pair* — `old → new` — and stays silently wrong when only one half is updated. Both #993 defects were exactly that:

- `147 → 150`: the branch merged `dev`, the baseline moved 147 → 149, and three of the four figures on the line were updated while `commands` was carried through. A **half-rebased delta**.
- `2,066 → 2,101`: the number was correct when measured, then a later commit changed the file and one of the two copies was updated.

Neither is a typo. Both are the predictable result of hand-editing a derived number in more than one place.

## Sketch of a fix

Roughly in order of value per effort:

1. **Add a delta pattern** — recognise `<counter> A → B` (and `A -> B`), then assert `B` equals the derived value. `A` can't be checked against the tree (it's a historical baseline), but asserting the *right-hand side* alone would have caught `147 → 150`.
2. **Widen file coverage** to `CHANGELOG.md` and all of `CLAUDE.md`, not just its two headline lines.
3. **Generate rather than validate.** The reviewer on #993 put it as: assert counters from a single generated source rather than copy-pasting across files. A `derive_counters.py --write` that rewrites the marked spans, with `--check` verifying the rendered output, removes the drift class instead of detecting it.

Option 1 alone closes the specific hole that shipped. Option 3 is the real fix and is a larger change.

## Prior art in this repo

The gate already grew twice in response to live escapes — the multi-occurrence scan was added after PR #985 shipped a README banner saying 380 while a section heading still said 370, and the badge and per-domain-table checks were added for the same reason. This is the same failure mode one level out: correct derivation, hand-typed prose.

## Scope note

Deliberately **not** fixed in #993. That PR was a skill compilation plus an emitter bug fix; changing the repo-wide counter gate inside it would have widened the diff well past its subject. Filing separately as agreed.

wakqasahmed:
## Summary

`derive_counters.py --check` gates absolute totals in five named files, but counter **prose** in `CHANGELOG.md` and `CLAUDE.md`'s release-note history sits entirely outside it — CHANGELOG.md was never read by this gate at all, and CLAUDE.md was only scanned on its two `**Current Scope:**`/`**Status:**` lines, not the release-note body where every past `Counters: skills N → M` entry actually lives. A fabricated delta (e.g. `commands 147 → 150` for a one-command change) passes CI green in both files, exactly as reported in #995.

## Fix

A one-directional invariant: since every one of these counters only grows over the repo's life, no delta claim's right-hand (new) value can legitimately exceed today's derived total. That's checkable without knowing which changelog entry is "current" — an equality check would fail on every legitimate historical entry describing a past, already-superseded state, which is most of what's in the file. `extract_delta_claims`/`check_delta_claims` implement this, scanning both `CHANGELOG.md` and the full text of `CLAUDE.md` (not just the two summary lines) for either arrow order ("skills 358 → 359" or "402 → 441 stdlib Python tools"), unicode or ASCII arrow.

## Testing

- `scripts/test_derive_counters.py` (new, stdlib unittest): the exact repro from #995 (both arrow styles) is rejected; a genuinely correct delta passes; an absolute/unchanged entry passes; a claim in a neighboring clause on the same line isn't cross-attributed to the wrong counter.
- Ran the actual repro against the real repo tree: appending the issue's fabricated `Counters: skills 999 → 1234; commands 12 → 999.` line to `CHANGELOG.md` now correctly fails `--check`; reverting it back to the clean tree passes clean (no false positives against the repo's own legitimate multi-year changelog history).
- Wired `test_derive_counters.py` into `ci-quality-gate.yml` (gate G3) so a future regression to the delta-claim logic itself gets caught, not just the original absolute-total check.

Closes #995.

---
_Implemented with AI assistance (Claude Code)._

### DIFF
--- a/.github/workflows/ci-quality-gate.yml
+++ b/.github/workflows/ci-quality-gate.yml
@@ -138,6 +138,10 @@ jobs:
         run: |
           python3 scripts/derive_counters.py --check
 
+      - name: Counter derivation unit tests (gate G3 — blocking)
+        run: |
+          python3 scripts/test_derive_counters.py
+
       - name: Safety dependency audit (requirements*.txt)
         run: |
           set -e
--- a/scripts/derive_counters.py
+++ b/scripts/derive_counters.py
@@ -25,7 +25,10 @@ Modes:
               disagree with derived values. Also validates the README
               "Skills Overview" per-domain table: every domain row's count
               must equal the SKILL.md count in its linked folder, and every
-              on-disk domain must have a row. CI gate G3.
+              on-disk domain must have a row. Also scans CHANGELOG.md and all
+              of CLAUDE.md for delta-style claims ("skills 358 -> 359") and
+              fails if any claimed new value exceeds the derived total (see
+              check_delta_claims). CI gate G3.
 
 Stdlib only. No writes ever.
 """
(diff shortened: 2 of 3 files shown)


wakqasahmed:
Hi @alirezarezvani — noticed this PR doesn't have a reviewer assigned yet — it's been sitting for a while, CI is green and it's mergeable. Would you (or whoever's best placed) be able to take a look, or point me to who should? Thanks!

wakqasahmed:
Hi @alirezarezvani — circling back on this one — still no reviewer assigned, quiet for a while since the last activity. No pressure, just don't want it to get lost.

PR #1012, opened 2026-09-06
write-a-skill: fix check 1 reading descriptions, and widen trigger vocabulary

### LINKED ISSUE #1011: [BUG] write-a-skill checklist runner: check 1 fails on `>-` descriptions and on "use whenever" phrasing
`write-a-skill`'s checklist runner has two defects in check 1 ("Description includes triggers") that make it report correct descriptions as failures. Both are still on `main` at `engineering/write-a-skill/skills/write-a-skill/scripts/skill_review_checklist_runner.py`.

I hit these while bringing a private plugin up to the 6-item standard, and measured them across 774 `SKILL.md` files: **33 skills fail check 1 while carrying a perfectly good trigger phrase.** A gate that fails correct work is worse than one that does not run, because it teaches people to ignore it.

## Defect A: block scalar indicators are compared by equality

`extract_frontmatter_description()` line 59:

```python
if val == ">" or val == "|":
```

YAML block scalar indicators may carry a chomping indicator (`>-`, `>+`, `|-`, `|+`) and/or an explicit indentation digit (`>2`). All are valid. Only the two bare forms are handled, so for any other variant the function falls through and returns **the indicator itself** as the description. Check 1 then compares trigger patterns against the literal string `">-"` and can never pass, whatever the description says.

```
description: >-
  Does a thing. Use when asked.
```
Check 1 fails. Change `>-` to `>` and it passes, with identical text.

This is easy to miss because Claude Code parses frontmatter with a real YAML parser, so the skill itself works fine. Only the checklist is wrong.

## Defect B: the trigger vocabulary is narrower than how people write

`TRIGGER_PATTERNS` is a fixed list of two-word strings. It accepts `use when` but rejects every one of these, all taken verbatim from shipped skills:

| written | why it fails |
|---|---|
| `Use whenever building a flow` | `\buse\s+when\b` needs a word boundary after `when`; `whenever` continues |
| `Use this whenever Matt says...` | filler word between verb and cue |
| `Use it whenever an audit is wanted` | same |
| `Use this skill whenever the user says...` | same |
| `This skill should be used when...` | `used`, not `use` |
| `Use on EVERY new store build` | cue is `on every` |
| `Invoke at the start of any session` | cue is `at the start of` |
| `Load this BEFORE writing or editing...` | verb is `load` |

Each of these is a clearer trigger than many that pass. The check is measuring phrasing, not presence.

## Proposed fix

Match an invocation verb followed by a temporal or conditional cue, rather than enumerating pairs. Both halves must still be present, so prose like "runs on macOS" or "teaches you to use design tokens" keeps failing.

```diff
-    if val == ">" or val == "|":
+    # A block scalar indicator may carry a chomping ("-", "+") and/or an explicit
+    # indentation digit: ">", ">-", ">+", "|2", "|-2" are all valid YAML.
+    if re.fullmatch(r"[>|][0-9]*[-+]?|[>|][-+]?[0-9]*", val):
```

```diff
-TRIGGER_PATTERNS = [
-    re.compile(r"\buse\s+when\b", re.IGNORECASE),
-    re.compile(r"\buse\s+for\b", re.IGNORECASE),
-    re.compile(r"\buse\s+before\b", re.IGNORECASE),
-    re.compile(r"\buse\s+during\b", re.IGNORECASE),
-    re.compile(r"\buse\s+after\b", re.IGNORECASE),
-    re.compile(r"\buse\s+while\b", re.IGNORECASE),
-    re.compile(r"\binvoke\s+when\b", re.IGNORECASE),
-    re.compile(r"\binvoke\s+before\b", re.IGNORECASE),
-    re.compile(r"\binvoke\s+after\b", re.IGNORECASE),
-    re.compile(r"\btrigger\s+when\b", re.IGNORECASE),
-    re.compile(r"\bapply\s+when\b", re.IGNORECASE),
-    re.compile(r"\brun\s+when\b", re.IGNORECASE),
-    re.compile(r"\brun\s+before\b", re.IGNORECASE),
-]
+_TRIGGER_VERB = r"(?:use|used|using|invoke|invoked|apply|applied|run|load|fire|trigger|reach\s+for)"
+_TRIGGER_FILLER = r"(?:\s+(?:this|it|its|the|a|an|skill))*"
+_TRIGGER_CUE = (
+    r"(?:when(?:ever)?|before|after|during|while"
+    r"|at\s+the\s+(?:start|point|moment)\s+of"
+    r"|on\s+(?:every|any|each))"
+)
+
+TRIGGER_PATTERNS = [
+    re.compile(rf"\b{_TRIGGER_VERB}\b{_TRIGGER_FILLER}\s+{_TRIGGER_CUE}\b", re.IGNORECASE),
+    # "Use for X" has no temporal cue but is long-standing and unambiguous.
+    re.compile(r"\buse\s+for\b", re.IGNORECASE),
+]
```

## Measured effect

Re-ran both versions over 774 `SKILL.md` files (this marketplace's plugins plus a private one):

```
now passing that used to fail        : 33
still failing (genuinely no trigger) : 140
regressions (passed before, fails now): 0
```

140 still fail, so the check keeps its teeth; it is not simply more permissive.

## Test

Twenty cases covering both defects, including four negatives that must keep failing:

```python
# every block scalar indicator must be read, not returned verbatim
('folded  >-',              block('>-', 'Does a thing. Use when asked.'), True),
('literal |-',              block('|-', 'Does a thing. Use when asked.'), True),
# real phrasings from shipped skills
('use whenever',            plain('Does a thing. Use whenever building a flow.'), True),
('should be used when',     plain('This skill should be used when the user says go.'), True),
('invoke at the start of',  plain('A router. Invoke at the start of any session.'), True),
('load this before',        plain('A reference. Load this before editing any schema.'), True),
# negatives: these must NOT start passing
('no trigger at all',       plain('Create developer handoff specifications with measurements.'), False),
('runs on macOS',           plain('A linter. Runs on macOS and Linux, written in Python.'), False),
('mentions use, no cue',    plain('Teaches you to use design tokens rather than raw values.'), False),
```

Happy to open a PR with the patch and the full test file if that is useful.

mattelms:
Fixes #1011.

Check 1 ("Description includes triggers") reports correct descriptions as failures, for two independent reasons. Both live in `skill_review_checklist_runner.py`, and `skill_description_validator.py` carries its own copy of each.

## 1. Block scalar indicators are compared by equality

`extract_frontmatter_description()` tests `val == ">" or val == "|"`. A YAML block scalar may also carry a chomping indicator or an explicit indentation digit, so `>-`, `>+`, `|-`, `|+` and `>2` all fall through and the function returns **the indicator itself** as the description. No description written that way can pass check 1, whatever it says.

```
description: >-
  Does a thing. Use when asked.
```

Check 1 fails. Change `>-` to `>` and it passes, with identical text.

`skill_description_validator.py` has the same bug in a milder form: `if val and val != ">"` leaves `>-` or `|` prepended to the value, which corrupts the length check and anything matched against the text.

This is easy to miss because Claude Code parses frontmatter with a real YAML parser, so the skills themselves work fine. Only the checks are wrong.

## 2. The trigger vocabulary is narrower than how people write

`TRIGGER_PATTERNS` is a fixed list of two-word strings. It accepts `use when` and rejects every one of these, each taken verbatim from a published skill:

| written | why it fails |
|---|---|
| `Use whenever building a flow` | `\buse\s+when\b` needs a word boundary after `when`; `whenever` continues |
| `Use this whenever Matt says run it` | filler word between verb and cue |
| `Use it whenever an audit is wanted` | same |
| `Use this skill whenever asked for hooks` | same |
| `This skill should be used when the user says go` | `used`, not `use` |
| `Use on EVERY new store build` | cue is `on every` |
| `Invoke at the start of any session` | cue is `at the start of` |
| `Load this BEFORE writing or editing a schema` | verb is `load` |

Several of these are clearer triggers than phrasings that already pass. The check is measuring phrasing, not presence.

## The fix

Match an invocation verb followed by a temporal or conditional cue, rather than enumerating pairs. Both halves must still be present, so prose like `runs on macOS` or `teaches you to use design tokens` remains a non-trigger.

## Measured on this repository's own skills

Both versions of the runner, over all 845 `SKILL.md` files here:

```
now pass, used to fail : 22
still fail             : 59
regressions            : 0
```

59 still fail, so the check keeps its teeth rather than simply becoming more permissive. Newly recognised include `landing`, `claude-coach`, `pw-init`, `pw-review`, `testrail`.

## Tests

Adds `engineering/write-a-skill/skills/write-a-skill/tests/`, following the `tests/`-beside-`scripts/` layout and `unittest` style of `engineering/skills/skill-tester/tests/`.

8 tests, 40 subtests: every block scalar indicator, the twelve phrasings above, five non-triggers that must keep failing, and one that asserts the two scripts' duplicated `TRIGGER_PATTERNS` reach the same verdict.

```
$ cd engineering/write-a-skill/skills/write-a-skill/tests
$ python3 -m unittest test_review_checklist_runner
Ran 8 tests in 0.001s
OK
```

Run against the unfixed scripts they produce 32 subtest failures, so they genuinely detect both defects rather than just passing alongside the fix.

The test file is force-added: `.gitignore` has a blanket `tests/` rule, and `engineering/skills/skill-tester/tests/` is tracked the same way, so this follows the existing precedent rather than changing your ignore rules.

## Consolidation (second commit)

The first commit fixed both copies and left them duplicated. That duplication is not incidental to the bug: it is the reason the bug had two different shapes, so the second commit consolidates it.

Three things were written more than once and disagreed:

| | |
|---|---|
| the trigger patterns | An identical copy in the runner and the description validator. Widening one left the other narrow. |
| the frontmatter reader | Each script had its own. The runner compared the indicator with `== ">"` and returned the indicator as the description. The validator tested `!= ">"` and left `>-` or `\|` prepended to the value. One defect, two shapes, because the reading was implemented twice. |
| the limits | 100 SKILL.md lines in two scripts, 1024 description characters in one, with nothing keeping them in step. |

All three now live in `scripts/_skill_frontmatter.py` - a module rather than a tool. The three scripts import from it and keep their own argparse entry points, so each is run exactly as before. They put their own directory on `sys.path` first, so they still work from any working directory.

**The cost, stated plainly:** this gives up the property that any one script can be copied out on its own. That is why I raised it in the original PR rather than assuming it, and I am happy to drop this commit if you would rather keep the scripts independent - the first commit stands alone.

**Behaviour-preserving, and not by assertion.** Every version was run over all 845 `SKILL.md` files here and compared verdict by verdict:

```
consolidated vs the duplicated fix, checklist runner       0 disagreements
consolidated vs the duplicated fix, description validator  0 disagreements
consolidated vs the duplicated fix, structure validator    0 disagreements
against upstream today       22 now pass, 59 still fail, 0 regressions
```

Tests go from 8 to 11. The new ones assert **identity rather than equality**, because two equal copies are exactly what drifted before: both scripts must hold the *same* `TRIGGER_PATTERNS` object, the same reader, the same `find_skill_md`. A fourth reads the scripts back and fails if any of them redefines a shared name, so a copy pasted in later is caught rather than merely discouraged.


mattelms:
Reopened as #1013 against `dev`, rebased onto it. My mistake targeting `main` here. #1013 also adds a second commit consolidating the duplicated trigger patterns, frontmatter reader and limits into one module.

